PR #6, opened 2016-02-02
Adding skeleton Jekyll site with placeholder theme

mtlynch:
This is just a placeholder site setup so that we can see our text copy rendered in Jekyll. Everything ending in .html or .scss will likely be completely rewritten/deleted when we design a real site.


critzo:
LGTM

PR #14, opened 2016-02-12
Initial stub out and structure of website conversion

shredtechular:
Apologize in advance for the number of files in this initial pull request, however, I tried to have this initial pull request to include bringing over several of the current site's assets (js, images, css, fonts, etc).  This also includes the stub out of most of the main landing pages (still need to add content to each, but at least contains the stub out to dynamically generate the navigational menu).

From a testing and viewing perspective, the header and footer common to all pages is complete.  Next Pull request will work on sub navigational items in the navigational header.  The 404 page is included in this pull request and is marked as complete on my list now.

You can pull this down and build locally and/or if you would like to see a preview of how the site is coming along and the features in this pull request, you can visit a temporary org/site I setup which will generate the site the this url: https://andrew-m-lab.github.io/

- [x] Initial README file that documents the organization of the site (what each directory includes) and the coding standards I'm trying to follow for the codebase.  Will add to this as I think of different standards to follow and open to anything folks would like to add here.  Plan is to keep up-to-date as I get further along in the conversion process.

- [x] The _config has the baseurl set to blank, which I think will be what is desired with how the main, parent repo will have the site structured.  This may need to change depending on where the site is desired to be generated, but simple enough to modify right in this config if needed.

- [x] Brings over key assets from current site to the Jekyll based site.  May have more to come here and will separate out any others as needed in future pull requests.

- [x] Stubs out the main landing pages with dummy, starter content to begin with.  Needed to do this to allow for dynamic generation of header navigation.

- [x] Header and footer common to all pages is complete with the exception of the dynamically generated sub navigational menu items, planning to do in next pull request which will stub out some more of those pages as well.

- [x] 404 page is included in this pull request and complete.

- [ ] You will notice that the RSS feed just links back to the home page.  Did this just to start as will need to include in another feature the generation of the RSS feed.  That pull request will also update the RSS feed links throughout the converted site.

PR #15, opened 2016-02-13
Addition of Observatory Page

shredtechular:
**Please NOTE** - This pull request builds upon https://github.com/m-lab/m-lab.github.io/pull/14 and should be merged in after that PR.

It includes the addition of the Observatory page which is mostly being injected via a git submodule into the project.

This was interesting to include and looks like it is working out just pretty well.  It can be previewed on my testing site here: https://andrew-m-lab.github.io/observatory/

Among some obvious items, in order to accomplish the use of submodules for this page, you will see:
- a new layout completely dedicated to this page (_layouts/observatory.html)
- updated the default layout **so the footer is not included on this page**
- updated the HEAD section for this page to include additional assets specifically for this page
- Finally, and most **importantly** I needed to pull the paths.js file out of the submodule directory and put it locally to the website.  I ran the ``python environment_bootstrap.py live`` command to create the symlink to the paths.js.live file, however it didn't appear I could check in symlinks, not I'm not sure we would want that regardless since the symlink lives in the submodule directory.  I think this solution works okay and is maintainable if the [mlab-observatory repo](https://github.com/m-lab/mlab-observatory) changes and the submodule in this site needs to be updated.
- additionally commanded travis to ignore the submodule directory as it was failing on the paths.js file since only paths.js.staging and paths.js.live exist in that directory.

- [x] Sidenote, also updated the README file to include a note about the submodule along with a few other corrections in there.


critzo:
LGTM @shredtechular

PR #17, opened 2016-02-13
Feature addition of Carousel & Homepage Elements

shredtechular:
**UPDATE**: Rebase completed for this PR.
~~**Please NOTE** - This pull request builds upon https://github.com/m-lab/m-lab.github.io/pull/14 and should be merged in after that PR.  There will most likely be **merge conflicts** after the first PR is accepted, so may need to update this PR once that is done.  There is several dependencies between many of these pages, so I think unavoidable as they all build on each other.~~

This update's major focus is on adding the carousel code as well as the other homepage functionality.  A preview of this functionality can be seen on my [staging site] (http://andrew-m-lab.github.io/).  Below are additional notes for the items completed with this pull request:

- [x] Carousel code is separated out into an include so that if it is decided to include a carousel on another page, one can add the frontmatter field ``carousel: true`` to the page and then add in values to the carousel as can be seen in home.md in the same format starting with ``carousel-items:``
- [x] New layout only for the homepage added as the layout is pretty unique to just the homepage.  This does **NOT** include any of the latest blog information on the current site.  This will be added when the blog posts are tackled to ensure it is pulling the latest blogs into that section.
- [x] In doing the conversion for the carousel/homepage, several other assets and pages that were not originally stubbed out are also now included so as not to break Travis builds not finding pages/links.

**UPDATE** - This will be done in a follow up pr, still good to merge this one.
~~- [ ] For several of these items, will need to go back in and ensure functionality in other pull requests is still working across the board, i.e. - may need to tweak breadcrumbs for pages like /tools/glasnost/~~

PR #18, opened 2016-02-14
Infrastructure Page

shredtechular:
**Please NOTE** - This pull request builds upon **ALL** previous pull requests made.  All pull requests should be merged in order.

This PR focuses on 
- adding the content for the infrastructure page
- js added in for https://github.com/m-lab/m-lab.github.io/pull/17 was breaking subnav functionality using pushstate.  I removed this as I saw a lot of funky behavior with the subnav on the current site.  If pushstate is absolutely desired for the navigation, we can revisit.  Most of the pages are loading pretty quickly, so not sure it is completely necessary in this circumstance.  If it is desired, I think it would be easier to work from a js file that wasn't modified if possible.


critzo:
@shredtechular LGTM except that we should add the subnav back in. If that's through a new js file, great. I'll go ahead with merging the outstanding PRs as is, and you can add the subnav in an upcoming PR.

PR #23, opened 2016-02-15
Addition of Blog Functionality

shredtechular:
**UPDATE** - This pull request has now been rebased on previous commits and is now ready for review when ready.
~~**Please NOTE** - This pull request builds upon previous PRs, specifically https://github.com/m-lab/m-lab.github.io/pull/22. Please don't merge until previous PRs are accepted and I've rebased.~~

This update provides the core blog functionality for the website conversion.  It does __NOT__ include ALL of the blog content.  The rest of the blog content migration will come in a future pull request, however, this feature does include all of the templates and partials needed for the base blog functionality.  It is conceivable a few tweaks may need to be made to the structure once the blog content migration is completed. This feature can be previewed [here](https://andrew-m-lab.github.io/blog/). I did make a few design decisions with the blog functionality conversion as its not a complete one-to-one identical match in all cases:

- [x] Constructed the blog index so that a user can page back and forth to older and newer blog entries.  Didn't see this ease of navigation on the current site.  If it is desired to take this out, we can remove it.  In order to implement this, I did add the ``jekyll-paginate`` gem, so you will see that update in the GemFile, associated configurations in the _config file, as well as using it in the blog index template.

- [x] The URLs of the current site use GET parameters, however, since Jekyll is static generated HTML files, I switched this so that a separate page is generated for each blog, blog index (including paging), and blog archives page.  A permalink was added to the config file to determine the URL structure.

PR #24, opened 2016-02-16
Content Migration - Performance Testing Tools Pages

shredtechular:
Minor pull request that adds the content in markdown format for the following performance testing tools pages:

- /neubot2/ and /neubot/  - Almost identical pages with the exception of some punctuation differences so put both in this PR even though /neubot/ is actually listed under the "Transparency" tools.
- /bismark/
- /pathload2/
- /sidestream/ - Also converted an inline style and put it in the base.css (conceivable this could cause a merge conflict with https://github.com/m-lab/m-lab.github.io/pull/23 but figured it would be good to get this PR in either way)

These pages can also be previewed [here](https://andrew-m-lab.github.io/tests/#performance).


critzo:
@shredtechular LGTM, but let's only have one page for Neubot since they are the same content. I'll merge this request and you can pull the /neubot2/ page out in a later PR, along with updating the links in the two different test sections so they both go to /neubot/

PR #29, opened 2016-02-17
Adding Publications Page & Associated PDFs

shredtechular:
This update includes the addition of the Publications Page which has links to a lot of associated pdfs, so a good majority of this PR is the addition of 30-some PDF assets to live locally with the site.

- [x] Converts all PDFs to live in a ``publications`` directory and have friendlier links as requested.  Did my best on the titles of the friendly links, but if there are any publication titles that should be altered, just let me know!

- [x] The biggest piece of this PR is the alteration of the "accordion" functionality so that the partial is more flexible and can handle multiple accordions on one page.  Previously, it was assuming there would only be one accordion per page.  I made some comments in the accordion partial on how to format different accordions in the markdown file and the publications post is a good example of how to do this.

A preview of this page can also be viewed [here](https://andrew-m-lab.github.io/publications/).


critzo:
@shredtechular LGTM. Merging.

PR #30, opened 2016-02-17
Fixing BISmark Reference

shredtechular:
Minor fix to correct a broken link and removes the ignore from Travis.


critzo:
@shredtechular LGTM. Merging.

PR #35, opened 2016-02-20
 Converting Additional Blogs & Fix for Pagination

shredtechular:
The majority of this update is for converting 3 additional blogs, however, in doing so, it also identified an issue with blog pagination, which this feature also fixes.

- [x] Adds the content, along with the corresponding category & monthly archives, for 3 more blog posts below:
    - /blog/internet_observatory_update/
    - /blog/2014_interconnection_report/
    - /blog/data_duplication_writeup/

- [x] Fixes a layout issue with the pagination.  The fix will now display the previous and next links to navigate blogs with each aligned right and left, instead of both aligned left previously.

- [x] Finally, noting that the javascript change is to add some formatting to the table on [this blog page](http://www.measurementlab.net/blog/data_duplication_writeup).  In order to maximize the use of markdown, I kept the table on this blog post all in markdown format, however in doing so, css was not going to do the job in formatting the red rows, so used js/jquery to format that row using ``$("tr:contains(Affected)").css('background-color','#ff5858').css('color', '#000000');`` to select all rows that have the word "Affected" it and applying the appropriate css 

Previewing of the newer blog entries can be viewed [here](http://andrew-m-lab.github.io/).


critzo:
@shredtechular LGTM. Merging.

PR #36, opened 2016-02-20
 Converting Additional Blogs & RSS Feed Functionality

shredtechular:
This update is for converting 2 additional blogs over to Jekyll.  One of these blogs had a link to the RSS feed, so I also added this to the pull request and updated links referencing RSS feeds across the site.

- [x] Adds the content, along with the corresponding category & monthly archives, for 2 more blog posts below:
    - /blog/observatory-new-features/
    - /blog/mlab-ripe/

- [x] Fixes a previously named blog post to match what is currently being used on the site.

- [x] Adds RSS feed functionality.  [GitHub Pages doesn't natively support traditional RSS feeds](https://help.github.com/articles/atom-rss-feeds-for-github-pages/) which is being used on the current site.  However, GitHub Pages does support Atom feeds which are newer than RSS 1.0 and look to be [widely used](https://github.com/jekyll/jekyll-feed/issues/2) through many RSS Feed Readers.  If it is desired to use traditional RSS feeds, then we could generate the RSS feed locally and check it into GitHub to serve.  This would mean though that anytime a blog is added that the RSS Feed would need to be regenerated, so I figured the GitHub Pages native solution would work best?

Previewing of the newer blog entries can be viewed [here](http://andrew-m-lab.github.io/) and the rss feed is [here](http://andrew-m-lab.github.io/feed.xml).


critzo:
@shredtechular LGTM. Merging.

PR #40, opened 2016-02-23
Content Migration - Contact Page

shredtechular:
This is a very minor and quick update to put in the content for the /contact page.  It was determined the contact form on the current site is not necessary to convert over, so this contains just the top paragraph of content.


critzo:
LGTM. Merging.

PR #42, opened 2016-02-24
Feature Homepage Latest Blogs Area

shredtechular:
I had forgotten about putting the "latest blogs area" into the homepage, right above the footer, so I added in some code into the homepage layout that dynamically gets the latest 2 blog posts and formats them into the homepage layout.  I also added some metadata into the home.md file that can control the content in the "Highlighted Visualization" section of that area.


critzo:
LGTM. Merging.

PR #43, opened 2016-02-24
Adding Google Analytics Script

shredtechular:
This pull request accomplishes the following:

- [x] - Noticed that google analytics was not previously included so adding that into the footer partial.

- [x] - Also noticed what looks like a script tag with embedded html that is supposed to be a fallback for mobile devices.  I wasn't able to see how this was implemented in the current site as browsing on my phone is not triggering that template to be shown.  I am adding that in however and in doing so needed to add an ignore-script-embed to the htmlproofer.  If we want to get rid of that script template then we can do that too...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ rvm:
   - 2.1
 script:
   - bundle exec jekyll build
-  - bundle exec htmlproof ./_site --only-4xx --check-html --file-ignore "/mlab_observatory/"
+  - bundle exec htmlproof ./_site --only-4xx --check-html --ignore-script-embeds --file-ignore "/mlab_observatory/"
 branches:
   only:
   - master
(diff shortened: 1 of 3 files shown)


critzo:
@shredtechular what do you need to see from.the current site re: analytics? Also, this branch is showing merge conflicts. Do we need to rebase and reissue the PR or manual merge? Thanks.

mtlynch:
@shredtechular - Is it possible to do a more granular suppression of htmlproofer so that it's only ignoring this one particular script rather than ignoring scripts globally?

shredtechular:
This PR has now been rebased and should be ready to get merged when ready.

@mtlynch I don't think it is possible to do a more granular suppression with htmlproofer for this scenario.  I attempted adding ignore tags to just that one embed script, but it still didn't pass, so I think you have to ignore script embeds gloabally https://github.com/gjtorikian/html-proofer#configuring-html-validation-rules to still get it to pass.

critzo:
LGTM. Merging.

PR #45, opened 2016-02-25
Removes CNAME file until we're ready.

nkinkade:
Removes CNAME file until we're ready... didn't realize Github would start doing auto redirects to custom domain.

### DIFF
--- a/CNAME
+++ b/CNAME
@@ -1,1 +1,0 @@
-www.measurementlab.net


mtlynch:
LGTM

For future PRs like these, consider git revert. I believe it creates a more obvious change history (though I'm not positive it would be meaningfully different in this case).

PR #46, opened 2016-02-26
Additional Documentation & Keywords Fix

shredtechular:
This pull request identified one bug fix as well as builds on some documentation needs in the README file and several include files.

- [x] In a final review of the conversion process, one other item was identified as needing a fix.  This fix was for the meta keywords tag.  On the current site, the keywords meta tag is only found on the home page, however, on the Jekyll site a mistake was hard coding the keywords meta tag and values into each page.  I moved the keywords information to the _pages/home.md file as a yml frontmatter field as well as added logic to the _includes/head.html file so that in the future if other pages need keyword values on the page, they can simply be added to the frontmatter for each page.  If there is no "keywords" value on the page, then it will not display the "keywords" meta tag.

- [x] Additionally, tweaked the README file adding in the publications directory to the structure and added some additional comments to the top of each include file to provide additional information if they need to be included to other pages in the future.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,7 @@ Current Build Status is: [![Build Status](https://secure.travis-ci.org/m-lab/m-l
 | fonts | Contains the customized font libraries for the project. |
 | js | Contains the js libraries for the project. |
 | images | Contains all the image files for the site |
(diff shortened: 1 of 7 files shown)


critzo:
LGTM. Merging.

PR #47, opened 2016-02-27
Feature addition of adding HTML Compression

shredtechular:
This pull request enables HTML compression for the mlab website.  Jekyll adds a significant amount of whitespace to the HTML when generating pages which can therefore add to the file size and degrade performance.  The performance gains by compressing HTML are probably not very noticeable to the avg end user, but doing some testing compressing the HTML vs. not compressing can be seen below.  Probably not an exact science but looks like on average about .4 KB size decrease and 30-40ms advantage.

mlab-optimizations


README.md
```diff
@@ -9,9 +9,13 @@ Current Build Status is: [![Build Status](https://secure.travis-ci.org/m-lab/m-l
 **Please Note** This repository contains a submodule, so after cloning this repo, you will also need to run `git submodule init` and `git submodule update` to pull down the submodule files as well.
 
 1. Install dependencies `bundle install`
-2. Run Jekyll server to preview in development mode `jekyll serve`.  To regenerate the site automatically use the watcher option `-w`.
+2. Run Jekyll server to preview in development mode `jekyll serve`.
 3. View the generated site by going to [http://localhost:4000/](http://localhost:4000/)
 
+### HTML Compression
+
+This site enables HTML Compression for optimizing performance.  If it is desired to not compress pages while doing development, developers can simply remove the ``layout: compress`` from the default template in the _layouts folder.  Please ensure to re-enable compression, however, before commiting code for review.
```
mtlynch: We can remove the last sentence as I think this kind of goes without saying.

shredtechular: OK, good call.  That sentence has now been removed.

critzo:
LGTM. Merging.

PR #2, opened 2016-01-25
Adding documentation for M-Lab's Google Cloud Storage data

mtlynch:
Rewriting documentation for M-Lab's GCS data. Content is a combination of:

* Original, added documentation
* https://github.com/m-lab/mlab-wikis/blob/master/HowToAccessMLabData.md
* https://github.com/m-lab/mlab-wikis/blob/master/ProjectHome.md

[Review on Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/2)

PR #1, opened 2016-01-25
Adding documentation for M-Lab's Google Cloud Storage data

mtlynch:
Rewriting documentation for M-Lab's GCS data. Content is a combination of:

* Original, added documentation
* https://github.com/m-lab/mlab-wikis/blob/master/HowToAccessMLabData.md
* https://github.com/m-lab/mlab-wikis/blob/master/ProjectHome.md

PR #48, opened 2016-02-27
Addition of yml validation and incorporation into CI

shredtechular:
**NOTE** - This will probably conflict with https://github.com/m-lab/m-lab.github.io/pull/47, so I can rebase after https://github.com/m-lab/m-lab.github.io/pull/47 is accepted/merged.

This pull request implements additional validation on yml used throughout the site and integrates it with Travis CI.

- [x] Implemented 2 forms of yml validation:
    - The first yml validation tests yml frontmatter contained in posts and pages against a yml "schema".  It is using a modified version of the jekyll_frontmatter_tests gem.  I needed to tweak the gem in order for the validation to skip nested directories, i.e. - _pages/categories, _pages/dates, etc.  So, this will enforce things such as pages or posts that REQUIRE certain yml frontmatter fields, formatting/linting of frontmatter, as well as yml key types, i.e. - strings, array.
**Example:** - Let's say a future developer leaves out the ``archive-type`` yml frontmatter in a category page and then ran this yml validation test, they would see something like below:
screen shot 2016-02-27 at 3 43 57 pm

    - The second yml validation tests basic linting of all yml files within the _data directory.  Currently there is only 1 file there, but if others are added it may be advantageous to have yml linting on those files too.  It would be possible to have a schema defined for these yml files as well, however, I'd only recommend this when the yml file is containing data other than just content.
**Example:** - Let's say a future developer incorporates a new yml file in _data and there is incorrectly formatted yml, running the yml validation test will display an error like below:
screen shot 2016-02-27 at 3 44 57 pm

- [x] In order to implement the yml validation within Travis CI, a grunt file was created with several tasks.  This should also aid in local development of the site too and can be expanded on if other tests/linting/tasks are desired in the future.  All of the grunt tasks created were also documented in the README.md file, but include:
    -  ``grunt serve`` task - intended for local development
    - ``grunt test`` task - intended to be run upon completion of development and in CI for tests to run
    - a default ``grunt`` task - intended to run the build and all tests, except htmlproofer since this one can take a little longer to execute all the external link checks.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ site/_site/
 coverage
 .ruby-version
 .sass-cache
+node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,16 @@
-language: ruby
-rvm:
-  - 2.1
+language: node_js
+node_js:
+  - 4.2
+before_install:
+# execute commands before dependencies can be installed
+  - rvm install 2.1  
+  - npm install -g grunt-cli
+install:
+# install dependencies
+  - bundle install
+  - npm install
 script:
-  - bundle exec jekyll build
-  - bundle exec htmlproof ./_site --only-4xx --check-html --ignore-script-embeds --file-ignore "/mlab_observatory/"
+  - grunt test
 branches:
   only:
   - master
--- a/Gemfile
+++ b/Gemfile
@@ -9,3 +9,7 @@ gem 'liquid'
 gem 'jekyll'
 gem 'jekyll-paginate'
 gem 'html-proofer'
+
+group :jekyll_plugins do
+	gem 'jekyll_frontmatter_tests', '0.0.6', :path => 'jekyll_frontmatter_tests-0.0.6'
+end
--- a/README.md
+++ b/README.md
@@ -8,9 +8,18 @@ Current Build Status is: [![Build Status](https://secure.travis-ci.org/m-lab/m-l
 
 **Please Note** This repository contains a submodule, so after cloning this repo, you will also need to run `git submodule init` and `git submodule update` to pull down the submodule files as well.
 
-1. Install dependencies `bundle install`
-2. Run Jekyll server to preview in development mode `jekyll serve`.
-3. View the generated site by going to [http://localhost:4000/](http://localhost:4000/)
+1. Install ruby dependencies `bundle install` (only needed upon initially cloning or when gemfile is updated)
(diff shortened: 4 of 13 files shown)


critzo:
@shredtechular #47 has been merged, so go ahead and rebase. Sorry for the delay.

shredtechular:
@critzo all set now.  Rebase was successful and should be okay to merge now.

mtlynch:
I very much appreciate automated validation/sanity checks, but I'm uncomfortable with this change because nobody on the team knows Ruby. We're not even comfortable in Jekyll, so my worry here is that if we change things about our schema in the future, we'll fail the validator and not really understand why or how to fix it.

shredtechular:
Hmm, that's a good point and I understand your concern with that.  Just to throw a couple of ideas out there to see if any of these would be a better way to attack it...

1. We could keep these checks in place, but just not force them to run with CI or local dev so future devs wouldn't necessarily get hung up on them?  They *could* be used if desired but if they aren't it wouldn't hold things up.

2. We could remove the first ruby/schema yml check entirely.  If you think it wouldn't be maintained and would only cause heartburn in the end, we could remove that check, but keep the yml linting grunt task?  That linter is just a node module ``grunt-yaml-validator``.  I don't know how advantageous it really is since there is currently only 1 file that linting check performs, but that could be an option if you'd like to go that route.

3. We could do some combination of #1 and #2 as well if you want one of the checks to be enforced but not the other, or keep 1 task but remove all from CI...

And I'm certainly okay too if we think we should just scrap this idea.  No reason to make things more complicated for any future work, the intention is to prevent issues/headache, but if this doesn't help, I'm in favor of just closing.

mtlynch:
I was interested in (2) but it feels very heavyweight to pull in grunt to do a YAML validation on a file we likely won't be editing that much. I don't think anyone on the team has experience with grunt and while I can kind of figure out what's going on in the code, I think we lose a lot of simplicity/readability compared to just being able to read the `.travis.yml` file and see the checks there.

@critzo - thoughts here?

critzo:
@mtlynch I think that your hesitation on both of these is sound. We have folks who have worked with Jekyll sites, but not much (if any) with grunt. We have other team members at OTI who have worked with Ruby, but I would rather not start using it now and require more capacity on the team that we don't really have now. I think we should pass on this PR and determine what remains to be finished before we publish the site (this week?).

mtlynch:
@shredtechular - Sorry we aren't able to use this. We really appreciate your efforts here to improve validation and I wish we had the in-house knowledge to be able to take advantage of this better.

PR #63, opened 2016-03-07
Bug - Broken Travis Builds

shredtechular:
I see that htmlproofer gem has been updated and therefore was causing some issues with the checks on travis builds.

In order to fix:
- updated some commands for the htmlproofer to work, using ``htmlproofer`` instead of ``htmlproof`` as well as removed the ``--ignore-script-embeds`` option now as it is deprecated.  By default script embeds are ignored unless specifically specified with the new htmlproofer option
- also updated rvm version to ensure latest environment and gems are used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: ruby
 rvm:
-  - 2.1
+  - 2.2.3
 script:
   - bundle exec jekyll build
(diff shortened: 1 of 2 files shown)


mtlynch:
Is it possible to specify an explicit version number of the gems we install so that we can't break if they update out from under us?

shredtechular:
@mtlynch good point.  I have specified the latest version for htmlproofer now, so any future bundle installs will make sure to use version 3.0.2.

mtlynch:
LGTM

PR #64, opened 2016-03-07
Bug - Fixing Baseurls

shredtechular:
This PR fixes the issue with baseurls to gain some consistency across local, forked repos, and the parent repo.  The solution decided upon in this pull request will make it so:
- For forked repos, no changes have to be made at all, ever
- For the parent repo, we are not using gh-pages functionality and it is deployed to s3.  Therefore, an sed utility command is being used to transform the baseurl to ``baseurl: ""``.  This ensures a good URL structure with the currently deployed website and makes sure the observatory page remains in working order.
- For local development, developers can use the ``--baseurl`` flag when generating the site locally.  This will ensure matching a production-like environment from a URL structure perspective.  The sed command should **NOT** be used locally as it overwrites the config file.

Finally, as part of this solution, there were some instances of the ``site.baseurl`` not being prepended in places it should have been that this PR also fixes.  When adding this in to certain places (specifically one link in the header, a few in the footer, and several on the /who page), additional logic needed to be added to ensure that absolute URLs and email references, i.e. - those with http, https, or mailto in them don't have the ``site.baseurl`` prepended to them.


mtlynch:
The ideal workflow is that someone can make a PR from their gh-pages branch, the reviewer can see what it looks like rendered, then that PR can be merged directly without any adjusting of config.yml.

>implementing something different with deployments to s3, but then the github pages wouldn't be doing anything.

I'd like to explore this one. We're actually using CloudFront+S3 to serve our site now, so measurementlab.net points there. Github pages is now useful mainly for development, but not for serving our production site.

shredtechular:
@mtlynch OK, got it that makes sense then.

If the parent repo is not using github pages at all and we don't need to support that, I added a new config file called "_config-deploy.yml" that is to only be used for the parent repo and deploying to s3.  This file will set the baseurl to "" and be used for the build that deploys to s3.  The existing _config can then be used for what actually gets built for forked repos and use the ``/m-lab.github.io`` baseurl.

Does that sound and work better from your perspective?

mtlynch:
Hmm, I'm not crazy about the idea of maintaining two mostly identical config.yml files.

Would it be possible to generate the deploy config dynamically in the _travis.yml script (e.g. using the `sed` utility to replace the `baseurl` field).

Does this get any easier if we just rename the repo to something like "github.com/m-lab/website"?

shredtechular:
Renaming the repo unfortunately wouldn't be the end-all cure as all the relative urls would have the ``/m-lab.github.io`` baseurl incorporated in the URL structure, so I imagine it would have unintended consequences when deploying to s3.

Will look into the sed utility feasibility to replace the baseurl field...

shredtechular:
OK, I added a baseurl.sed script into a new _scripts directory as well as updated the travis file to include executing that script and then moving the dynamically generated _config.yml overwriting the _config in the root before travis builds, checks, and deploys the generated site.

That is the first time for me using the sed utility (awesome to learn a new trick!) so please let me know any feedback you may have for any better implementation methods.

If you are good with this solution, then I will update the documentation in the README file, just want to finalize the solution before updating that if it's okay.

_scripts/baseurl.sed
```diff
@@ -0,0 +1,5 @@
+# Sets the baseurl to an empty string to be used before deploying to s3
+# command to implement only for the _config.yml file is:
+# sed -f _scripts/baseurl.sed _config.yml > _scripts/_config.yml
+
+s/\/m-lab\.github\.io//g;
```
mtlynch: Can we do this inline instead of having a separate script file? Also, would prefer if we used a regex to match the entire line (including baseurl) to reduce the chances of this step clobbering `m-lab.github.io` that might appear on another line.

.travis.yml
```diff
@@ -1,6 +1,9 @@
 language: ruby
 rvm:
   - 2.2.3
+before_script:
+  - sed -f _scripts/baseurl.sed _config.yml > _scripts/_config.yml
+  - mv _scripts/_config.yml _config.yml
```
mtlynch: I actually liked the name `_config_deploy.yml` so it's clear to the reader why we're doing this. I also think it's useful for the reader to see in the `jekyll build` call that it's using `config_deploy.yml` explicitly.

nkinkade: To Michael's other comment about whether to do this inline or have a sed file, I agree that doing it inline is probably better.  I think you can remove the "mv" line and replace the sed line with:

```sed -ie '/baseurl/s/\/m-lab\.github\.io//' _config.yml```

The -i option make the change on the file in-place, so no need to redirect the output.  And adding /baseurl/ to the expression assures us that this operation will _only_ happen on the line containing "baseurl", just in case we ever, for some reason, have another entry or line with the same string (unlikely, but good form).

Comment (mtlynch) [inline .travis.yml]: @nkinkade - How do you feel about giving it a separate name? I think writing to a new file called `_config_deploy.yml` makes it a little clearer what's going on. Overwriting a source-controlled file also feels a little dirty to me.

Comment (nkinkade) [inline .travis.yml]: Would a comment in _config.yml suffice instead of writing a new file? It seems like a comment would be less crufty and vastly more explicit than the implicit message sent by redirecting output to file with "deploy" in the name.

Comment (mtlynch) [inline .travis.yml]: Comment is fine. I don't feel very strongly about separate filename.

Comment (shredtechular) [inline .travis.yml]: Excellent feedback, thank you guys!

I have updated this PR to address the latest comments to inline the sed replacement of the baseurl and it now overwrites the existing _config.yml file without a ``mv``.  Thank you for showing me how to do that!

I also added a couple comments into the travis, config, and README files that hopefully will make it explicitly clear what is going on.

**Comment by mtlynch** [comment]
Cool, can you submit for formal review?

File: .travis.yml [hunk 3]
```diff
@@ -1,6 +1,8 @@
 language: ruby
 rvm:
   - 2.2.3
+before_script:
+  - sed -ie '/baseurl/s/\/m-lab\.github\.io//' _config.yml # sets baseurl to "" in _config for deploying w/o unnecessary url paths
```
mtlynch: Can we put the comment on the previous line? I'd like to avoid exceeding 80 columns when possible.

_config.yml
```diff
@@ -1,9 +1,13 @@
 # Site settings
 title: M-Lab
 email: support@measurementlab.net
-baseurl: ""
 excerpt_separator: <!--more-->
 
+# Default baseurl used for gh-pages branches.
+# Travis will change it to "" before deploying.
+# For local development pass in blank baseurl to simulate prod-like environment: `jekyll serve --baseurl`
```
mtlynch: Suggest formatting like

```
...environment:
$ jekyll serve --baseurl
```

mtlynch:
Can you update the PR description to describe the ultimate solution we settled on.

Also can you include details in the PR description about the changes to the template logic? I understand the places where we're prefixing {{ baseurl }} but I'm not clear on what's going on with the link checks or the checks for mailto vs. http.

shredtechular:
PR description has now been updated describing the finalized solution and the reason for the mailto vs. http logic.  If any of that needs further clarification, let me know.

.travis.yml
shredtechular: Absolutely, good catch.

This comment is now moved to the previous line and changed to suggested wording.

Also added the caret to the regex as suggested to match only at the beginning of a string.

_config.yml
shredtechular: Fixed the formatting as per suggested.

mtlynch:
>When adding this in to certain places (specifically one link in the header, a few in the footer, and several on the /who page), additional logic needed to be added to ensure that absolute URLs and email references, i.e. - those with http, https, or mailto in them don't have the site.baseurl prepended to them.

I prefer to whitelist rather than blacklist when possible. Can we switch the logic so that instead of prepending *unless* we see `http` or `mailto`, we *only* prepend if the url starts with a forward slash?

shredtechular:
OK, that makes sense.  The PR has been updated to whitelist rather than blacklist and look for a forward slash as the first character.

mtlynch:
LGTM

PR #76, opened 2016-03-09
Embed NDT using scripts from ndt-project, but HTML from this repo.

nkinkade:
We don't want to maintain yet another implementation of a Websocket test to run NDT tests, yet we can't use the code in the [official NDT repo](https://github.com/ndt-project/ndt/tree/master/HTML5-frontend) out of the box because it is fairly specific to being used in the fakewww environment.  This PR makes the necessary changes to our website repository to make this happens, and hinges on a [number of other changes](https://github.com/ndt-project/ndt/pull/204) to the core NDT Websocket code.

This PR could be simplified by removing the added fonts and selecting some other ones that browsers will already have installed.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/76)


_layouts/ndt-test.html
```diff
@@ -2,33 +2,95 @@
...
-  }
-};
-xhr.open("GET", mlabNsUrl, true);
-xhr.send();
-</script>
+    </style>
+    <![endif]-->
+    <script>
+        var ndtScriptPath = "{{ site.baseurl }}/ndt/HTML5-frontend/";
+        var CANV_GAUGE_FONTS_PATH = "../fonts/digital-7-mono";
```
mtlynch: Can we document what these variables are doing?

css/ndt-test.css
```diff
@@ -0,0 +1,337 @@
...
+
+
+/* Test */
+
+#test .locations {
+  width: 640px;
+  position: absolute;
+  left: 40px;
+  top: 400px;
+  /*color: white;*/
```
mtlynch: Can we delete this?

css/ndt-test.css
```diff
@@ -0,0 +1,337 @@
+@font-face {
```
mtlynch: It seems like we're inconsistent about how we're using blank lines. I don't have a preference either way, just as long as we're consistent.

css/ndt-test.css
```diff
@@ -0,0 +1,337 @@
...
+#results .view-selector {
+  position: absolute;
+  right: 40px;
+  top: 60px;
+  font-size: 18px;
+  line-height: 18px;
+}
+#results .view-selector .summary {
+  border-width: 2px 1px 2px 2px;
+  padding-right: 15px;
```
mtlynch: Should just be `padding`
https://google.github.io/styleguide/htmlcssguide.xml?showone=Shorthand_Properties#Shorthand_Properties

(ditto for 318-319, 326)

_layouts/ndt-test.html
```diff
@@ -2,33 +2,95 @@
...
+                    <p>
+                        <span class="number" id="upload-speed">0.00</span>
+                        <span class="units" id="upload-speed-units">mb/s</span>
+                    </p>
+                </div>
+                <div class="download result">
+                    <h3>Download speed</h3>
+                    <p>
+                        <span class="number" id="download-speed">0.00</span>
+                        <span class="units" id="download-speed-units">mb/s</span>
```
mtlynch: Should be Mb/s

mtlynch:
It looks like Travis is failing due to non-UTF-8 characters in the source.

css/ndt-test.css
nkinkade: I'm not sure that this should be reduced to shorthand. The declaration is only for right padding... he (Aaron) apparently didn't want top, left or bottom padding for some reason.  If I tried to make that into shorthand, I'd be overriding the inherited padding values for those other sides.

The same goes for 318-319, where trying something like "padding: 0 15px" would set the left and right padding to 15px, but would also set top padding to 0, which it wouldn't have otherwise been (probably).

If I'm incorrect, could you provide some examples of how to do this without wiping out inherited values for the other sides?

css/ndt-test.css
nkinkade: I think this should be fixed now with commit @75e3b5a.

ndt
```diff
@@ -0,0 +1 @@
+Subproject commit e4aa3380f3a2c76b75de5822f23a91d38ddc0ecf
```
mtlynch: Should be `0e5585173c566f50260a84e6976804a569bb36ea`

_layouts/ndt-test.html
```diff
@@ -2,33 +2,116 @@
...
+        // variable tells it where to look.
+        var CANV_GAUGE_FONTS_PATH = "../fonts/digital-7-mono";
+    </script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/jquery-1.12.1.min.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/gauge.min.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/script.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/ndt-browser-client.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/ndt-wrapper.js"></script>
+    <script>
+        // Lookup a suitable NDT server using mlab-ns. The variable ndtServer is
```
mtlynch: (nit) Should be "Look up" since "lookup" is a noun.

_layouts/ndt-test.html
```diff
@@ -2,33 +2,116 @@
...
+        // used by script.js in the ndt submodule
+        var ndtServer;
+        var mlabNsService = 'https:' == location.protocol ? 'ndt_ssl' : 'ndt';
+        var mlabNsUrl = 'https://mlab-ns.appspot.com/';
+        $.ajax({
```
mtlynch: Is there a race condition here? What if the user tries to start the test before we have a result from mlab-ns?

_layouts/ndt-test.html
```diff
@@ -2,33 +2,116 @@
...
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/gauge.min.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/script.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/ndt-browser-client.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/ndt-wrapper.js"></script>
+    <script>
+        // Lookup a suitable NDT server using mlab-ns. The variable ndtServer is
+        // used by script.js in the ndt submodule
+        var ndtServer;
+        var mlabNsService = 'https:' == location.protocol ? 'ndt_ssl' : 'ndt';
+        var mlabNsUrl = 'https://mlab-ns.appspot.com/';
```
mtlynch: The globals on 31 and 32 seem to violate the rule against globals. https://google.github.io/styleguide/javascriptguide.xml?showone=Naming#Naming

_layouts/ndt-test.html
```diff
@@ -2,33 +2,116 @@
...
+        $.ajax({
+            url: mlabNsUrl + mlabNsService + '?format=json',
+            dataType: 'json',
+            async: false,
+            success: function(resp) {
+                console.log('Using NDT server: ' + resp.fqdn);
+                ndtServer = resp.fqdn;
+            },
+            error: function(resp) {
+                console.log('mlab-ns lookup failed.');
```
mtlynch: Can we log anything more specific?

css/ndt-test.css
```diff
@@ -0,0 +1,366 @@
...
+  font-family: "League Gothic";
+  src: url("../fonts/league-gothic/league_gothic.otf") format("opentype");
+}
+
+body * {
+  margin: 0;
+  padding: 0;
+}
+
+body {
```
mtlynch: Doesn't this make 6-9 redundant?

css/ndt-test.css
```diff
@@ -0,0 +1,366 @@
...
+  position: absolute;
+  right: 40px;
+  top: 60px;
+  font-size: 18px;
+  line-height: 18px;
+}
+
+#results .view-selector .summary {
+  border-width: 2px 1px 2px 2px;
+  padding-right: 15px;
```
mtlynch: >I'm not sure that this should be reduced to shorthand. The declaration is only for right padding... he (Aaron) apparently didn't want top, left or bottom padding for some reason. If I tried to make that into shorthand, I'd be overriding the inherited padding values for those other sides.

>The same goes for 318-319, where trying something like "padding: 0 15px" would set the left and right padding to 15px, but would also set top padding to 0, which it wouldn't have otherwise been (probably).

>If I'm incorrect, could you provide some examples of how to do this without wiping out inherited values for the other sides?

My guess is that the fact that we can't let particular attributes slip through is intentional by the guide. But the wording is vague enough ("whenever possible") that I'm not sure if it needs to be observed here and I don't want to spin on this.

_layouts/ndt-test.html
nkinkade: I don't think so.  To avoid this is why I've set "async: false" in the $.ajax() request. My understanding is that this should block the entire page until the call finishes.  I think the reason you are seeing that error is because the front-end code changes, but the ndt submodule was still pointing to the outdated generic-js branch of the m-lab/ndt repo.  I've updated the submodule and expect it to work after I push the commits to the PR in a minute.

_layouts/ndt-test.html
nkinkade: Issue address in @b2cc2da

_layouts/ndt-test.html
nkinkade: Addressed in @b2cc2da, I believe.

_layouts/ndt-test.html
nkinkade: Addressed in @b2cc2da.

css/ndt-test.css
nkinkade: I don't think so, since what goes in "body {" applies to the body itself, whereas what goes in "body * {" applies specifically to children of the body element. Though it could be that margin and padding are both inherited by all children anyway?

ndt
nkinkade: Addressed in @c2469d8

css/ndt-test.css
mtlynch: My understanding is that they inherit. Can we delete 6-9 and see if you get the same behavior? If so, leave deleted, otherwise keep.

_layouts/ndt-test.html
```diff
@@ -2,33 +2,120 @@
...
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/gauge.min.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/script.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/ndt-browser-client.js"></script>
+    <script src="{{ site.baseurl }}/ndt/HTML5-frontend/ndt-wrapper.js"></script>
+    <script>
+        // Look up a suitable NDT server using mlab-ns. The global variable
+        // ndtServer is used by by the backend javascript to determine which
+        // server to test against.
+        var ndtServer;
+        setMlabServer();
```
mtlynch: Kind of confusing to call a function we haven't defined yet. Suggest making it an anonymous function:

```
(function() {
  var mlabNsService = 'https:' == location.protocol ? 'ndt_ssl' : 'ndt';
  var mlabNsUrl = 'https://mlab-ns.appspot.com/';
  ...
})();
```

Described here: http://esbueno.noahstokes.com/post/77292606977/self-executing-anonymous-functions-or-how-to-write

_layouts/ndt-test.html
```diff
@@ -2,33 +2,120 @@
...
+        function setMlabServer() {
+            var mlabNsService = 'https:' == location.protocol ? 'ndt_ssl' : 'ndt';
+            var mlabNsUrl = 'https://mlab-ns.appspot.com/';
+            $.ajax({
+                url: mlabNsUrl + mlabNsService + '?format=json',
+                dataType: 'json',
+                async: false,
```
mtlynch: > To avoid this is why I've set "async: false" in the $.ajax() request. My understanding is that this should block the entire page until the call finishes. 

Ah, can we add a comment above this line saying it's deliberately false so that the page doesn't load until we get the result from mlab-ns?

_layouts/ndt-test.html
```diff
@@ -2,33 +2,120 @@
...
+                success: function(resp) {
+                    console.log('Using NDT server: ' + resp.fqdn);
+                    ndtServer = resp.fqdn;
+                },
+                error: function(jqXHR, errStatus, errText) {
+                    console.log('mlab-ns lookup failed: ' + errStatus + errText);
```
mtlynch: Do we need a space between? Like `errStatus + ' ' + errText`?

mtlynch:
LGTM

nkinkade:
I added fixes for you final round of comments, Michael... the site builds fine on Travis and the deployment the gh-pages branch of my fork look good.  Going to merge.

Issue #74, opened 2016-03-09
Formatting on the test pages for links to data is incorrect


georgiamoon:
The bullets aren't showing in the Data section of some of the pages:

https://www.measurementlab.net/tools/ndt/
https://www.measurementlab.net/tools/npad/
https://www.measurementlab.net/tools/paris_traceroute/

example from NDT:
> Data collected by NDT is available…
> in raw format at https://storage.cloud.google.com/m-lab/ndt
> via an SQL interface (see https://github.com/m-lab/mlab-wikis/blob/master/BigQueryMLabDataset.md).
> Visualized in Public Data Explorer.
> Please cite this dataset as follows: The MLab NDT Dataset, <date range used>. http://measurementlab.net/tools/ndt

Overall, the formatting is a little easier to read on OONI Probe & Glasnost, but I'll wait to see the CSS changes that @shredtechular is working on before I file more issues :)

PR #79, opened 2016-03-11
Feature: Initial Setup of SASS capabilities & Data Pg Styles

shredtechular:
This PR contains the setup of moving towards SASS as well as some styles and components specifically related to the Data Pages styles.  Also, note that this PR is being made against the new-data-docs branch instead of master.

The changes for the data pages as well as any other pages can be viewed on my [forked site](http://shredtechular.github.io/m-lab.github.io/data/gcs/).

The ideology behind the structure of the SASS files that were used is below:

**Directory** - Description 
**base** - Includes boilerplate elements and typography styles 
**components** - Includes specific modules that are common to multiple pages
**helpers** - Includes helper functions, mixins, and variables 
**layouts** - Will contain the main sections of the layout, i.e - header, footer, body, etc. 
**pages** - Don't have anything here yet, but this directory is meant to contain any pages that may have abnormal styling for some reason that may go against common elements 

Further notes and clarifications:
- The main components that were put into a SASS format were lists, code syntax highlighting, and tables.  Left other elements and styles as is in the base.scss file which compiles those styles as is.  The goal is to all other styles into SASS, but for sake of biting off small pieces, this is where the refactor of the css begins.

- Some design decisions were made to simplify the css and provide consistent styling across the site.  For example, there were several different ways in which lists were showing across the site, as @georgiamoon pointed out in https://github.com/m-lab/m-lab.github.io/issues/74.  As she points out, there are pages on the current site that use lists of type disc, lists of type square, lists of type circle, and lists that don't have bullets at all.  This refactor begins to consolidate some of these styles and use them more consistently throughout the site.  For main content areas, I went with disc lists, unless one of the following occurs: (A) a nested list in which case the next level nesting will show circle type lists, or (B) lists contained within accordion content will also show as circle type lists.  We could conceivably treat ALL lists as type circle, unless specifically pointed out to be different within the markdown, but thought this was the best way to go to begin with.

- Another bigger design decision made was for the code syntax highlighting.  I saw some SASS being used in an early example/prototype, so incorporated some of the design elements into this refactor.  Most noticeably, areas where code syntax is being used, it will highlight the whole area to make it stand out a little bit more.  I am open to how this looks to folks and can alter or remove if folks don't like it.

- Stubbed out the variables.scss page however did not begin refactoring styles to use this just yet.  I think once further refactoring is in place, we will be able to take advantage of this more, but wanted to get an initial PR in to move the data pages styles forward.

- Added compression for when the SASS is compiled to CSS.

- During the refactor, there was one occurrence of a specific class that I removed from a markdown file as it was not needed if we are going for a more consistent representation of lists.

- Also during the refactor, there was one blog post that had code tags used, however, the syntax highlighting is not going to work well with this example due to the width of the code example, so created a no-highlight class so it is not affected.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/79)


critzo:
LGTM @mtlynch @georgiamoon Perhaps you should also review this one?

mtlynch:
LGTM

Issue #75, opened 2016-03-09
Add Travis check for trailing whitespace


mtlynch:
We should add a Travis check to make sure that files in this repo do not have trailing whitespace (at the very least .md and .yml files).

This can either a simple grep for "\s+$" or can be a more thorough check using a Markdown linter (bonus points if we can smoothly integrate a markdown linter into Travis/git hooks, especially one that reflows lines to break at 80 columns)

mtlynch:
@shredtechular can you take this on after tackling the style issues?

shredtechular:
@mtlynch Absolutely, I can see what I can figure out that may help with that respect after the styling fixes are complete.

PR #80, opened 2016-03-12
Feature: Markdown Linter

shredtechular:
This PR integrates the markdown linter found [here](https://github.com/mivok/markdownlint).  It has several rules that can be configured based on our needs and how much we want to follow some of the markdown best practices they have rules around.  This is one solution around https://github.com/m-lab/m-lab.github.io/issues/75.

**Please note**, we should carefully consider adding this linter.  I don't want to create extra burden on authors trying to meet these markdown linting rules, but if the team feels it would be beneficial, then I think it's a good decision.  We can remove it if folks think it is too burdensome.  **If the only goal is to prevent whitespace**, then I have also put together a pre-commit git hook that can run locally as well as in Travis that may work better.  We could also configure this linter to **ONLY** fail if whitespace is found if desired.

**Regardless** if the linter is something we want, I think we could still work in this PR with some of the markdown fixes that were made.  We could just take out the linter if it is **NOT** wanted.

**Finally** if this linter is implemented, we will also need to run this on the new-data-docs branch eventually.  This PR was branched off of master, so doing the PR against that branch.

In order to implement this solution, the following actions were taken:

- [x] Updated the Gemfile to include the markdown linter gem

- [x] Setup a markdown linter configuration that disables and customizes a few rules to best work for our purposes in ``_test/markdown-linter-style.rb``.  Also updated Travis to be able to run the linter on the README file, the _posts directory, and the _pages directory.  These 3 commands could be simplified down to 1 command to work on the whole project, however, it did find issues in the submodule's README.  If we updated the submodule's README to pass the linting rules then we can change these commands.

- [x] Once the linter was in place, I also fixed a number of issues that it found in several markdown files.  These fixes included lots of whitespace at the ends of lines, removal of several consecutive bland lines, extra spaces when defining list items, one occurrence of an incorrect HR rule being defined, and actually it caught an unneeded div in the ndt tool page.

Here is what the linter errors look like either in Travis or the same can be done locally:
screen shot 2016-03-12 at 11 25 47 am

---
This change is [Review on Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/80)


.travis.yml
```diff
@@ -2,10 +2,14 @@ language: ruby
 rvm:
   - 2.2.3
 before_script:
+  # Markdown Linter using custom rule style file.
+  - mdl README.md --style _test/markdown-linter-style.rb
+  - mdl _posts --style _test/markdown-linter-style.rb
+  - mdl _pages --style _test/markdown-linter-style.rb
```
mtlynch: Can we just run this on everything in `./`?

_pages/tools/windrider.md
```diff
@@ -7,7 +7,7 @@ breadcrumb: tests
 
 # Windrider
 
-**Windrider has been decommissioned on 01/17/2013. The source code will be still available at the link provided below.**
+Windrider has been decommissioned on 01/17/2013. The source code will be still available at the link provided below.
```
mtlynch: Why are we changing the formatting here?

_test/markdown-linter-style.rb
```diff
@@ -0,0 +1,11 @@
...
+exclude_rule 'MD001'
+rule 'MD009', :br_spaces => 2
+exclude_rule 'MD013'
+exclude_rule 'MD022'
+exclude_rule 'MD026'
+rule 'MD029', :style => :ordered
+exclude_rule 'MD032'
+exclude_rule 'MD033'
+exclude_rule 'MD034'
+exclude_rule 'MD036'
```
mtlynch: Can we document what these rules are with comments?

_test/markdown-linter-style.rb
```diff
@@ -0,0 +1,11 @@
+all
```
mtlynch: Would prefer the pathname to be something like _config or _linter

README.md
```diff
@@ -1,20 +1,29 @@
 # [Measurement Lab](http://www.measurementlab.net/) Source Code
 
-This is the source code of the Measurement Lab website built using [Jekyll](http://jekyllrb.com) and utilizing [GitHub Pages](https://pages.github.com/) to publish and host the site.
+This is the source code of the Measurement Lab website built using
```
mtlynch: Why are we breaking lines at 80 columns in this file but not others?

.travis.yml
shredtechular: Per PR description, yes, we could simplify those 3 commands down to 1 command on ``./`` however, the linter did find issues in the m-lab observatory submodule's README. If we updated the submodule's README to pass the linting rules then we can change these commands.  I didn't see an easy way to just ignore that external submodule's directory, so that is why I opted for the 3 commands.

_pages/tools/windrider.md
shredtechular: Sorry, should've pointed this one out, good catch.

The linter identified having an entire paragraph bolded as an issue, stating it would be best practice to use a heading level instead of bolding an entire paragraph.  

I notice that on the /tools/pathload2/ page, it also has a similar "decomissioned" statement, but it is an h2.  Do you think it would be better to just change this to an h2 so it is consistent with the other pages where they have a decommissioned sentence?

README.md
shredtechular: OK, I can revert these changes.  I started down the path of doing that and decided if we wanted to go that route no reason to back those changes out, but I can do that.

_pages/tools/windrider.md
shredtechular: This has been reverted now.

_test/markdown-linter-style.rb
shredtechular: Comments added in https://github.com/shredtechular/m-lab.github.io/commit/5ac4e73b2716cce1f79a34e5219c6dbac4d91e6c

_test/markdown-linter-style.rb
shredtechular: Changed the pathname to be _linter instead of _test in https://github.com/shredtechular/m-lab.github.io/commit/5ac4e73b2716cce1f79a34e5219c6dbac4d91e6c

README.md
shredtechular: Reverted line reflowing in readme in https://github.com/shredtechular/m-lab.github.io/commit/1bd89ca17e269f20c670ec48ac9ad4090416561a

_linter/markdown-linter-style.rb
```diff
@@ -0,0 +1,33 @@
...
+all
+
+#  Rule Modifications
+#  --------------------------------
+## MD029 mod - tells linter not to fail if order list items don't start with 1
+rule 'MD029', :style => :ordered
+
+#  Exclude these rules
+#  --------------------------------
+## MD001 - Check that Header levels increment by one level at a time.
```
mtlynch: Can we reorganize this so that there's a list of rules we are excluding by design and then another list of rules we're excluding *for now* because our current content does not adhere to the rule? It looks like everything except for "line length" and "inline HTML" are rules we actually want to obey.

_linter/markdown-linter-style.rb
```diff
@@ -0,0 +1,33 @@
+#  Include all rules by default
+all
+
+#  Rule Modifications
+#  --------------------------------
+## MD029 mod - tells linter not to fail if order list items don't start with 1
```
mtlynch: This seems like a rule we'd want to follow. Do we have ordered lists that don't begin with 1?

.travis.yml
```diff
@@ -2,10 +2,14 @@ language: ruby
 rvm:
   - 2.2.3
 before_script:
+  # Markdown Linter using custom rule style file.
+  - mdl README.md --style _linter/markdown-linter-style.rb
+  - mdl _posts --style _linter/markdown-linter-style.rb
+  - mdl _pages --style _linter/markdown-linter-style.rb
```
mtlynch: > Per PR description, yes, we could simplify those 3 commands down to 1 command on ./ however, the linter did find issues in the m-lab observatory submodule's README. If we updated the submodule's README to pass the linting rules then we can change these commands. I didn't see an easy way to just ignore that external submodule's directory, so that is why I opted for the 3 commands.

Ah, bummer. Other two ideas:

1. Does `--git-recurse` recurse into git submodules? If not, then that would work.
1. Can we specify multiple targets per command (e.g. `mdl README.md _posts _pages --style _linter/markdown-linter-style.rb`)?

_linter/markdown-linter-style.rb
shredtechular: Maybe the comment isn't very clear here, was trying to think of a way to succinctly say it...

Basically, it tells the linter that it is okay to have ordered lists that increment, which seems like it should be the default, but the default for the linter is it is expecting each ordered list item to start with 1.  For example:

By default the linter expects ordered list items to be like below:

    1. Do this.
    1. Do that.
    1. Done.

However, we have list items like below that this configuration then allows::

    1. Do this.
    2. Do that.
    3. Done.

shredtechular: OK, I added a more descriptive comment now that hopefully explains this modification better.

_linter/markdown-linter-style.rb
shredtechular: Absolutely, I re-organized these rules according to by design and until content/code is re-worked.  I recommend we keep the "bare url" rule (MD034) as being excluded by default -- I don't really agree with that MD check in general and not sure if we would get it to work nicely with yml frontmatter if we ever wanted a URL in there.

Also, added issues for the rules we are temporarily (hopefully) excluding.

.travis.yml
shredtechular: Yes, your #2 suggestion worked perfectly!  Why didn't I think of that! :+1: 

Fixed in https://github.com/shredtechular/m-lab.github.io/commit/b2d5335c3e6d0e473081d9356bbf12d902864e45

_linter/markdown-linter-style.rb
```diff
@@ -0,0 +1,41 @@
+#  Include all rules by default
+all
+
+#  Rule Modifications
+#  ----------------------------------------------------------------------------
+## MD029 - tells linter to allow ordered lists that will increment by 1.  By 
+## default (w/o modification) the linter doesn't accept lists structured like:
+## 1. List item #1
+## 2. List item #2
+## 3. List item #3
```
mtlynch: Ah, I get it. I actually prefer the 

```
1.
1.
1.
```

style because you don't have to manually re-number when you delete items, but I think this might be a controversial position on the team, so I'm fine with leaving this as is for now.

_linter/markdown-linter-style.rb
```diff
@@ -0,0 +1,41 @@
...
+rule 'MD029', :style => :ordered
+
+
+#  Exclude these rules by design
```
mtlynch: Suggest expanding a bit more like, "Exclude the following linter rules that conflict with the team's Markdown writing preferences."

mtlynch:
One request for comment clarification but LGTM

_linter/markdown-linter-style.rb
shredtechular: Sounds good to me.

Fixed in https://github.com/m-lab/m-lab.github.io/pull/80/commits/8324b1f7179a8ed42d03876bdf31b80080af28d8

PR #81, opened 2016-03-12
Feature: Pre-commit hook to execute Travis checks

shredtechular:
This PR adds a pre-commit git hook that can be installed locally by using the following command: ``rm -rf .git/hooks/ && ln -s -f ../_hooks .git/hooks``.

The way it works is whenever a developer/author is working locally and attempts to run a ``git commit`` command, the hook fires and will perform the same checks that are run on Travis (minus check external links due to speed).  Currently, this means that it will run the markdown linter and htmlproofer.

During implementation of this pre-commit git hook, the following items should be noted:

- In order to optimize maintainability of the checks that Travis as well as now the git precommit hook are executing, a design decision was made to create a separate file that will house all of the checks that either one has to do.  This means any future maintenance of these checks can be simply be updated in one file.

- In order to do all the travis checks in the hook, the sed config alteration also needed to be done. I didn't like how it was altering the _config.yml file in place as this could cause issues with accidentally committing unintended changes to a very important file.  Therefore, a design decision was made to alter the sed config alteration to create a _config-deploy.yml file.  I also added that config to the gitignore so we aren't changing files under source control and no one has a chance of accidentally committing it. 

- If we are simply executing the travis checks in the pre-commit hook, then we are relying on the markdown linter for whitespace checking.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/81)


mtlynch:
I think this is actually a bit more complex a solution than we need.

Some high level notes:

1. I don't think a pre-commit should change code that's about to be committed, even if it does alert the user. I'd rather it just fail with an informative error message (i.e. which lines are bad). Bonus if it fails and then shows the user a command they can copy/paste to fix it themselves, but failing with informative error messages is sufficient.
1. In general, we shouldn't have complex bash logic in our code. One liners are fine, but when it gets to the point of conditionals and loops, we should consider alternatives. Python is preferred since every developer on the team understands Python. I don't recommend Python here because I think we can make do without complex logic in this pre-commit, but just for future reference in general.
1. In terms of file naming/layout/documentation, I should have mentioned earlier, but try to model after what we do with hooks in bigsanity: https://github.com/m-lab/bigsanity/tree/master/hooks

shredtechular:
OK, I simplified the pre-commit hook based on the feedback and am ONLY checking to see if whitespace is introduced by simple git diff check.  If it fails this check, then it outputs the file/line info detailing where trailing whitespace was entered as well as an sed command that can be used to fix the issues by file.

Also added some bash formatting to try and make it clear to end users if the commit doesn't work due to this whitespace check as can be seen below in a test commit locally:

screen shot 2016-03-14 at 3 54 17 pm

_hooks/pre-commit
```diff
@@ -0,0 +1,22 @@
...
+#
+
+# A git hook script that checks for trailing whitespace
+# in your commits. Bypass it with the --no-verify option
+# to git-commit
+#
+# usage: make a soft link to this file,
+# e.g., ln -s ~/config/pre-commit.git.sh ~/some_project/.git/hooks/pre-commit
+
+if test "$(git diff --check --cached)"
```
mtlynch: Can we instead use the same checks that Travis is doing?

_hooks/pre-commit
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+#
+
+# A git hook script that checks for trailing whitespace
```
mtlynch: Can we structure this the same as bigsanity (same example command in README, similar wording)

https://github.com/m-lab/bigsanity/tree/master/hooks

shredtechular: Ah, I see what you mean with that now, yes, we can do the hooks that way too, I'll fix that.

_hooks/pre-commit
shredtechular: I'm not sure I understand what you mean here?  Do you mean to incorporate this hook into Travis as well and fail if it finds whitespace?

_hooks/pre-commit
shredtechular: OK, structured the hook the same as bigsanity with one exception.  It appears as though having the hooks in a dir called ``hooks`` and then creating a symlink to it is causing jekyll to spit out errors when building locally as opened under the jekyll-watch gem here: https://github.com/jekyll/jekyll-watch/issues/36

I recommend we keep the hooks in ``_hooks`` instead of ``hooks`` for the time being until it is possible to exclude dirs that have symlinks from the jekyll-watcher.

README.md
```diff
@@ -12,9 +12,13 @@ Current Build Status is: [![Build Status](https://secure.travis-ci.org/m-lab/m-l
 2. Run Jekyll server and pass in a blank baseurl to preview in development mode `jekyll serve --baseurl`.
 3. View the generated site by going to [http://localhost:4000/](http://localhost:4000/)
 
+### Pre-commit Hook
+
+Local Developers should also install the pre-commit hook that comes packaged with this repository that will alert about any trailing whitespaces to minimize git diff noise.  In order to install the pre-commit hook, run the following command to create a symbolic link: ``rm -rf .git/hooks/ && ln -s -f ../_hooks .git/``.
```
mtlynch: Why the double backtick? I've only ever seen single and triple.

_hooks/pre-commit
mtlynch: Oh, sorry I should have elaborated more. Basically, can we make the pre-commit hook identical to the Travis check, but maybe a little more abbreviated? So something like:

```
mdl README.md _posts _pages --style _linter/markdown-linter-style.rb && \
  bundle exec jekyll build && \
  bundle exec htmlproofer ./_site --only-4xx --check-html --file-ignore "/mlab_observatory/" \
    --disable_external
```

I'd like the pre-commit hook to complete in <15s, so that's why I'd like to remove the external check. But the goal is that users don't have to do a PR just to see if the build checks are passing and can get reasonably strong confidence just from the pre-commit hook.

_hooks/pre-commit
mtlynch: >I recommend we keep the hooks in _hooks instead of hooks for the time being until it is possible to exclude dirs that have symlinks from the jekyll-watcher.

Yeah, this sounds good to me.

_hooks/pre-commit
shredtechular: Ok, I understand what you are saying here now.  You'd like the pre-commit hook to not only check whitespace, but also run the other checks that are currently being run with travis and can be run locally, but with individual commands.  Let me see what I can stir up on this piece...

README.md
shredtechular: Good catch, changed that to triple now as that seems like that has benefits over single, but don't think it matters in this case.

Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/2eab78f260fb67321df8d52aa62a71143f1c37ef

_hooks/pre-commit
shredtechular: OK, I have incorporated the travis checks into this hook now as well as rebased from master to pull in the latest changes.  In redoing the hook code to pull in all of the travis checks there are a couple of consequences:
1. I didn't like hardcoding the same travis commands in the pre-commit hook AND the .travis file, so I created a new file called _tests/travis-checks.sh that can then be run in the hook & on travis and only has to be maintained in one spot.  Since the hook is not to check external links, but it should on travis, the hook will pass in an argument to disable external link checks.
2. In order to do all the travis checks in the hook, the sed config alteration also needed to be done.  I didn't like how it was altering the _config.yml file in place, so I think it would be better to now alter it to a _config-deploy.yml file so we aren't changing files under source control and noone has a chance of accidentally commiting it.  I added the _config-deploy.yml to the .gitignore to prevent any commit issues with that file too.
3. Since we took out the ws check, we are relying on the markdown linter for whitespace checking.  I didn't add any instructions for fixing ws issues since it could really fail for any number of reasons.  If we want to grep on the issues and add instructions to fixing whitespace issues ONLY IF found in the grep, we can do that, but adds more complexity to the bash script which I think we were trying to avoid.

README.md
```diff
@@ -12,6 +12,10 @@ Current Build Status is: [![Build Status](https://secure.travis-ci.org/m-lab/m-l
 2. Run Jekyll server and pass in a blank baseurl to preview in development mode `jekyll serve --baseurl`.
 3. View the generated site by going to [http://localhost:4000/](http://localhost:4000/)
 
+### Pre-commit Hook
+
+Local Developers should also install the pre-commit hook that comes packaged with this repository that will alert about any trailing whitespaces to minimize git diff noise.  In order to install the pre-commit hook, run the following command to create a symbolic link: ```rm -rf .git/hooks/ && ln -s -f ../_hooks .git/hooks/```.
```
mtlynch: Suggest either cutting "Local" (what's a non-local developer?) or lowercasing "Developers".

_hooks/pre-commit
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+#
+
```
mtlynch: Extraneous newline

_hooks/pre-commit
```diff
@@ -0,0 +1,21 @@
...
+# A git hook script that runs the same checks that are
```
mtlynch: Can we break these lines at 80 columns?

_hooks/pre-commit
```diff
@@ -0,0 +1,21 @@
...
+# usage: make a symlink to this file,
+# e.g., rm -rf .git/hooks/ && ln -s -f ../_hooks .git/hooks
+
+# Run 
+./_tests/travis-checks.sh --disable-external
+
+if [ $? -eq 0 ]; then
+    exit 0
+else
+  echo "**********************************************************"
```
mtlynch: Suggest using an `err()` function:

https://google.github.io/styleguide/shell.xml?showone=STDOUT_vs_STDERR#STDOUT_vs_STDERR

_hooks/pre-commit
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
mtlynch: Just a general comment: we use the Google style guidelines for most languages, so for this one, we'll want to adhere to the shell style guide:

https://google.github.io/styleguide/shell.xml

_hooks/pre-commit
```diff
@@ -0,0 +1,21 @@
...
+# e.g., rm -rf .git/hooks/ && ln -s -f ../_hooks .git/hooks
+
+# Run 
+./_tests/travis-checks.sh --disable-external
+
+if [ $? -eq 0 ]; then
+    exit 0
+else
+  echo "**********************************************************"
+  echo -e "*  \033[31mCOMMIT ABORTED! Please address the errors before commiting.\033[0m"
```
mtlynch: I appreciate the nice formatting, but I think it makes the code harder to maintain. I'd prefer to just have a simple error message with no colored fonts or star separators.

_hooks/pre-commit
```diff
@@ -0,0 +1,21 @@
...
+#
+
+# A git hook script that runs the same checks that are 
+# executed on Travis to verify your commits will not fail.
+# Bypass it with the --no-verify option to git-commit
+#
+# usage: make a symlink to this file,
+# e.g., rm -rf .git/hooks/ && ln -s -f ../_hooks .git/hooks
+
+# Run
```
mtlynch: Suggest deleting this comment as it doesn't provide much information

_tests/travis-checks.sh
```diff
@@ -0,0 +1,15 @@
...
+mdl README.md _posts _pages --style _linter/markdown-linter-style.rb
+
+# Github Pages needs the baseurl field to be "m-lab.github.io" to render properly in forked 
+# repos, but when we deploy to production, we need the baseurl field to be empty. This 
+# command replaces "m-lab.github.io" with an empty string.
+sed -e '/^baseurl/s/\/m-lab\.github\.io//' _config.yml > _config-deploy.yml
+
+# Run HTMLProofer
+bundle exec jekyll build --config _config-deploy.yml
+bundle exec htmlproofer ./_site --only-4xx --check-html --file-ignore "/mlab_observatory/","/ndt/" $1
```
mtlynch: Should be `"$1"`:
https://google.github.io/styleguide/shell.xml?showone=Quoting#Quoting

Instead of doing a blind passthrough of the first parameter, suggest making the parameter to travis-checks.sh like `--quick` and then if we see `--quick` passed to `travis-checks.sh`, we add `--disable-external` to this command.

mtlynch:
I like this strategy. Can we update the PR description so that it matches the current implementation?

README.md
shredtechular: Good point, removed "Local".

Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/3f6000223f6819c18a5133a60e77369a825dff00

_hooks/pre-commit
shredtechular: Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/3f6000223f6819c18a5133a60e77369a825dff00

_hooks/pre-commit
shredtechular: Re-formatted comment to break closer to 80 columns.

Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/3f6000223f6819c18a5133a60e77369a825dff00

_hooks/pre-commit
shredtechular: OK, that works.

Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/3f6000223f6819c18a5133a60e77369a825dff00

_hooks/pre-commit
shredtechular: OK, removed error message formatting and added ```err()``` function.  Didn't comment this function as it seemed small an obvious, but if you prefer a comment for this function I can add one.

Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/3f6000223f6819c18a5133a60e77369a825dff00

_hooks/pre-commit
shredtechular: Agreed, good eyes.  Removed that comment.

Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/3f6000223f6819c18a5133a60e77369a825dff00

_tests/travis-checks.sh
shredtechular: OK, I like this suggestion.  Made the argument ```--quick``` now and then in the travis-checks file it looks for the first argument and replaces "quick" with "disable-external" if found.  If it doesn't find "quick" then it will not execute HTMLProofer with any other flags.  Could also do this with a more traditional if statement I suppose, but seemed like this implementation was a little more succinct.  If you had something else in mind we can change that piece.

For now marking this as fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/3f6000223f6819c18a5133a60e77369a825dff00

_hooks/pre-commit
shredtechular: OK, good to know.  I did also see that in the guidelines it is strongly preferred not to use ```.sh``` extension, so removed that from the travis-checks file.  Will try to follow that more in the future on any bash work.

shredtechular:
OK, PR description has now been updated to match the current implementation and all edits made for previous comments.

_hooks/pre-commit
```diff
@@ -0,0 +1,26 @@
...
+
+#######################################
+# Error function to display error message to end users.
+# Arguments:
+#   Error message description
+#######################################
+err() {
+  echo "$@"
+  exit 1
+}
```
mtlynch: Oh sorry, I actually meant we should wholesale steal the example from here:
https://google.github.io/styleguide/shell.xml?showone=STDOUT_vs_STDERR#STDOUT_vs_STDERR

Primarily because it prints to stderr instead of stdout. Also I'd prefer not to bake the "exit" into the function.

_tests/travis-checks
```diff
@@ -0,0 +1,18 @@
...
+
+# Execute Markdown Linter
+mdl README.md _posts _pages --style _linter/markdown-linter-style.rb
+
+# Github Pages needs the baseurl field to be "m-lab.github.io" to render properly in forked 
+# repos, but when we deploy to production, we need the baseurl field to be empty. This 
+# command replaces "m-lab.github.io" with an empty string.
+sed -e '/^baseurl/s/\/m-lab\.github\.io//' _config.yml > _config-deploy.yml
+
+# sets --disable-external flag for HTMLProofer if quick argument exists
```
mtlynch: Prefer to just go more generic and say "skips validation checks that take a long time" because we will likely have other checks in the future that we want to run quickly in the pre-commit, but go more exhaustively in Travis

_tests/travis-checks
```diff
@@ -0,0 +1,18 @@
...
+quick="${1/%quick/disable-external}"
```
mtlynch: Prefer to just make this an `if`/`then` where we compare `$1` to `--quick`

_hooks/pre-commit
```diff
@@ -0,0 +1,26 @@
...
+# git commit
+#
+# usage: make a symlink to this file,
+# e.g., rm -rf .git/hooks/ && ln -s -f ../_hooks .git/hooks
+
+#######################################
+# Error function to display error message to end users.
+# Arguments:
+#   Error message description
+#######################################
```
mtlynch: Oh we don't need a function docstring here. If we reduce it to a one-liner, it's self-explanatory.

README.md
```diff
@@ -12,6 +12,10 @@ Current Build Status is: [![Build Status](https://secure.travis-ci.org/m-lab/m-l
 2. Run Jekyll server and pass in a blank baseurl to preview in development mode `jekyll serve --baseurl`.
 3. View the generated site by going to [http://localhost:4000/](http://localhost:4000/)
 
+### Pre-commit Hook
+
+Developers should also install the pre-commit hook that comes packaged with this repository that will alert about any trailing whitespaces to minimize git diff noise.  In order to install the pre-commit hook, run the following command to create a symbolic link: ```rm -rf .git/hooks/ && ln -s -f ../_hooks .git/hooks/```.
```
mtlynch: I think as a general rule we use single backticks for inline code, triple backtick for code blocks. Suggest switching to single backtick unless we have a specific reason not to.

_hooks/pre-commit
shredtechular: Ah, okay, that makes sense, okay, updated that now to the `err()` function verbatim.

Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/eabd7618abc1915efa0d31b50e1c9b29e56e6078

_tests/travis-checks
shredtechular: Right, good call.

Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/eabd7618abc1915efa0d31b50e1c9b29e56e6078

_tests/travis-checks
shredtechular: OK, switched this to an if statement now.  I was going back and forth about whether or not it would best to separate that into a function now since that logic is becoming longer.  I kept it simple for now, but if you think additional logic could eventually go in here then maybe we consider a function for this?  Maybe that's overkill...

Implemented an `if` statement now though at least in https://github.com/m-lab/m-lab.github.io/pull/81/commits/eabd7618abc1915efa0d31b50e1c9b29e56e6078

_hooks/pre-commit
shredtechular: Dang, I coulda swore I took that out as I decided it was overkill, shoulda looked the files over again after committing.

Fixed now in https://github.com/m-lab/m-lab.github.io/pull/81/commits/eabd7618abc1915efa0d31b50e1c9b29e56e6078

README.md
shredtechular: OK, that make sense.

Fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/eabd7618abc1915efa0d31b50e1c9b29e56e6078 along with some other instances in the README where inline code had triple backticks.

_tests/travis-checks
```diff
@@ -0,0 +1,21 @@
...
+# Execute Markdown Linter
+mdl README.md _posts _pages --style _linter/markdown-linter-style.rb
+
+# Github Pages needs the baseurl field to be "m-lab.github.io" to render properly in forked 
+# repos, but when we deploy to production, we need the baseurl field to be empty. This 
+# command replaces "m-lab.github.io" with an empty string.
+sed -e '/^baseurl/s/\/m-lab\.github\.io//' _config.yml > _config-deploy.yml
+
+# Skips validation checks that take a long time
+arg=""
```
mtlynch: Suggest `htmlproofer_args_extra` so that it sets us up well for the future, like if we have a separate command we want to add/change arguments to when we see the `--quick` switch.

mtlynch:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 17 unresolved discussions, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/81#-:-KD4kLanxDtJYMPUb4Zm:1417088721)*

_tests/travis-checks
shredtechular: OK, that makes sense to me.

That is fixed in https://github.com/m-lab/m-lab.github.io/pull/81/commits/31ec265f36f15bb9efa7af643d1e90d111da55c8

mtlynch:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 17 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/81#-:-KD51APPAYXux2yoSDh-:1417088721)*

PR #83, opened 2016-03-13
Adding Anchors for Data Pages

shredtechular:
This PR adds quick link anchors at the top of 2 data pages per [issue 77](https://github.com/m-lab/m-lab.github.io/issues/77).

I wasn't sure if anchors were desired on any of the other pages?  I went with the logic that any of these pages that had a couple H1 headers that is what we have been putting quick link anchors on, so continuing with that theme.  If we want quick link anchors on the /data/bq/quickstart/ page, should we consider changing some of the main section headers to H1s or do we want to change the use of these?  It seems like it would give a more consistent experience for the end user to expect H1s for these anchors?  I'm open to what you all think is best though.

That aside, I was able to add anchors for the /data/gcs/ and /data/bq/examples/ pages.  Those changes can be seen on my fork [here](http://shredtechular.github.io/m-lab.github.io/data/gcs/) and [here](http://shredtechular.github.io/m-lab.github.io/data/bq/examples/), respectively.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/83)

### DIFF
--- a/_pages/bigquery_examples.md
+++ b/_pages/bigquery_examples.md
@@ -3,12 +3,24 @@ layout: page
 title: "BigQuery Examples"
 permalink: /data/bq/examples/
 breadcrumb: data
+accordion-quick-links: true
+quick-links-section:
+  - column:
+    - group-heading: "Overview"
+    - group-heading: "Basic counting"
+    - group-heading: "Computing statistics over time"
+  - column:
+    - group-heading: "Dealing with IP addresses"
+    - group-heading: "Comparing NDT and NPAD tests"
+    - group-heading: "Computing distributions of tests across users"
 ---
 
+{:.overview}
 # Overview
 
 The examples below query the M-Lab data in various ways to demonstrate effective use of the M-Lab BigQuery dataset.
 
+{:.basic-counting}
 # Basic counting — How many users?
 
 Let's start with something simple. How many distinct users (distinct IPs for simplicity) have ever run an **NDT** test?
@@ -28,6 +40,7 @@ Result:
 |-------------|
 | 110814220   |
 
+{:.computing-statistics-over-time}
 # Computing statistics over time — How many users per day?
 
 By slightly modifying the previous query, it is possible to compute how the number of users changed over time.
@@ -62,6 +75,7 @@ ORDER BY
 | 2015-12-27  |       52995 |
 | 2015-12-28  |       50751 |
 
+{:.dealing-with-ip-addresses}
 # Dealing with IP addresses — How many users from distinct subnets?
 
 BigQuery supports various functions to parse IP addresses in different formats. You can use such functions to aggregate the number of users per subnet and compute how many subnets have ever initiated a test.
@@ -83,6 +97,7 @@ FROM
 |-------------|
 | 4548859     |
 
+{:.comparing-ndt-and-npad-tests}
 # Comparing NDT and NPAD tests — How many users have run both NDT and NPAD tests?
 
 * The following query computes the number of distinct IP addresses that have run tests using both NDT and NPAD.
@@ -114,6 +129,7 @@ FROM
 |----------------|
 |           74535|
 
+{:.computing-distributions-of-tests-across-users}
 # Computing distributions of tests across users — How many users have run a certain number of tests?
 
 Now let's try something a bit more complex.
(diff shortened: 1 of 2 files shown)


critzo:
For broad accessibility, we should not use multiple h1 tags on any page. I would be in favor of a solution that allows easily adding IDs to any H[123456] tag.

shredtechular:
I think the most important thing from an accessibility perspective is that heading levels are being used semantically and that the heading levels are being used in order, right?  For example we shouldn't be going from an h1 to an h3.  What's the reason for concern for multiple h1s on an html5 page?

As for the solution though, any heading level can be used for quick links, my point was more from a consistency across the site where the "quick link" anchors are showing.  If we are going to display these quick links, it seems like we would want them used in a consistent way across the site.  So, for example, on this page, h1s were used: /data/gcs/ so I added quick links to that page.  However, on this page: /data/bq/quickstart/ h2s were used (skipping the h1 heading levels) so I did NOT use quick link anchor tags.

critzo:
The H1 tag is usually the title of the page or the top level subject of the
page, so semantically and logically only one H1 tag on a page should be
needed. I've always used a single H1 followed by properly nested sub
headings.

@georgiamoon @mtlynch do you have thoughts on this?

 Chris Ritzo
 Senior Technologist, Open Technology Institute @ New America
 740 15th Street NW, Suite 900
 Washington, DC 20036
 Public Key:    0x085B309C

mtlynch:
`h1` for title sounds fine to me. I don't mind adding an anchor link to h1s, especially if we can do it like Github does default Markdown formatting where the anchor link only appears when the user hovers over the text, like in:

https://github.com/m-lab/telescope/blob/master/documentation/selector-file-spec.md#example-file

georgiamoon:
A few thoughts, because I think I wasn't clear before, so wanted to address that.

1) On the GCS and BQ Examples page, those probably should be h1s as @critzo has pointed out, I believe from an accessibility compliance perspective, then need to be h2s.

2) My thought was that there would be a way to programmtically create anchor links for any subsections, e.g. that that could be something that happens based on the markdown of a header existing, but it looks to me like this is a manual change? Can someone clarify?

georgiamoon:
Also -- the I think the anchor links should update in the URL when clicked, and not just be a hidden link.

critzo:
+1 @georgiabullen. I like the ability to add a sort of table of contents at the top of some pages like:

But those "accordion" links should also update the URL. 

Also it looks like we can add an id tag to any element (H tags or others) using this syntax:
 +{:.name-of-id/anchor} 

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/83#-:-KCq3_jgVEoMwi0e1kc-:-1059175490)*

shredtechular:
RE: ``2) My thought was that there would be a way to programmtically create anchor links for any subsections, e.g. that that could be something that happens based on the markdown of a header existing, but it looks to me like this is a manual change? Can someone clarify?``

Currently, we do need to specify anchor links in the yml frontmatter of a page if we want a page to display anchor links at the top of the page, which gives a decent amount of flexibility for different pages of the site.  We can do this programmatically, but if we do that, we need to define the business rule behind this.  For example, across all pages on the site, do we make a business rule that any page that has multiple h1s have anchor links at the top of the page?  Or perhaps, we define a css class to add to any headers that should include an anchor link?  Then we could target with js and create the table of content/anchor links at the top of the page?


RE: ``Also -- the I think the anchor links should update in the URL when clicked, and not just be a hidden link.``

Agreed, however, this should be a sitewide change I think and wonder if it should be in its own PR?  This PR was focused on [issue 77](https://github.com/m-lab/m-lab.github.io/issues/77) to just ensure anchor links were added to the top of the data  pages.

shredtechular:
RE: ``Also it looks like we can add an id tag to any element (H tags or others) using this syntax:
+{:.name-of-id/anchor}``

This actually adds a class to those elements as that is what the existing js was using for anchor links at the top of the page.  If we go to a more natural implementation of hash tag anchor links, then we shouldn't need to add those classes.  We would still need to identify the anchors that need the Table of Content links though in the yml frontmatter or come up with a business rule for when a Table of Content/anchor links should show at the top of a page.

critzo:
This PR was superceded by PR #119 
Closing.

Issue #85, opened 2016-03-14
BigQuery Schema Page: Table style issues


mtlynch:
On the BigQuery Schema page, there are a few style issues. See here:

https://mtlynch.github.io/m-lab.github.io/data/bq/legacymigration/

1. The tables are too wide and use an unnecessary amount of empty space
1. The tables don't have sufficient padding/margins separating them from subsequent text:
![image](https://cloud.githubusercontent.com/assets/7783288/13753475/7c0e72c0-e9e9-11e5-8b00-e1405df3baa9.png)

(1) might be an issue that's unique to this page, so it's fine if the change only applies to that page. (2) is probably something we want to fix globally.

PR #91, opened 2016-03-14
Bug - Data Table Styling

shredtechular:
This PR is for some tables styling fixes that @mtlynch identified in https://github.com/m-lab/m-lab.github.io/issues/85.  You can see the fixes in action on my fork [here](http://shredtechular.github.io/m-lab.github.io/data/bq/schema/).

I think the fixes that can be applied to fix the styling on the data pages should both be more global fixes, however, there is a table in a blog page that is requiring the min-width attribute to be applied to one of its headers to ensure the formatting stays nice on that page too.  That's why I added the ``big-query`` class to that table in the blog post.

---
This change is [Review on Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/91)

### DIFF
--- a/_posts/blog/2015-04-24-data_duplication_writeup.md
+++ b/_posts/blog/2015-04-24-data_duplication_writeup.md
@@ -26,7 +26,7 @@ This issue was introduced into the dataset sometime in May 2014. We fixed the ro
 
 The table below details the effect of the duplication bug on each of the BigQuery tables. Note that the table is organized not by time of query, but according to the time the M-Lab test occurred, as that is how M-Lab data is partitioned into tables in BigQuery.
 
-{:.big-query}
+{:#bigquery-schema-fields}
 |---
 | Month / BigQuery Table | Impact
 |:-|:-
--- a/_posts/blog/2015-04-24-data_duplication_writeup.md
+++ b/_posts/blog/2015-04-24-data_duplication_writeup.md
@@ -26,6 +26,7 @@ This issue was introduced into the dataset sometime in May 2014. We fixed the ro
 
 The table below details the effect of the duplication bug on each of the BigQuery tables. Note that the table is organized not by time of query, but according to the time the M-Lab test occurred, as that is how M-Lab data is partitioned into tables in BigQuery.
 
+{:.big-query}
 |---
 | Month / BigQuery Table | Impact
 |:-|:-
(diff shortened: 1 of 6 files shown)


critzo:
@shredtechular LGTM
@mtlynch Do you want to take a quick look to confirm this solves the issue as you submitted?

_posts/blog/2015-04-24-data_duplication_writeup.md
```diff
@@ -26,6 +26,7 @@ This issue was introduced into the dataset sometime in May 2014. We fixed the ro
 
 The table below details the effect of the duplication bug on each of the BigQuery tables. Note that the table is organized not by time of query, but according to the time the M-Lab test occurred, as that is how M-Lab data is partitioned into tables in BigQuery.
 
+{:.big-query}
```
mtlynch: Suggest `bigquery` since BigQuery is one word

_sass/pages/_blog.scss
```diff
@@ -0,0 +1,7 @@
+// Unique elements for blog pages
+
+.blog-page {
+  table.big-query {
+    @include tables(200px);
```
mtlynch: I don't think it makes sense to support the "BigQuery" table on blog pages because I don't anticipate a use for that.

mtlynch:
Looks much better. LGTM.

_posts/blog/2015-04-24-data_duplication_writeup.md
shredtechular: I like making it an id instead of a class, you make a good point there.  I am hoping the base styles for tables should be more flexible moving forward, so I don't think we need anything special at the moment for large tables.  It probably just depends on how much bigger some future tables get though, so we may need to revisit in the future.

shredtechular: Comment addressed.  Fixed in https://github.com/m-lab/m-lab.github.io/pull/91/commits/cfd1724768b3c1f78fd772dc279e3cde618b3d6b.

_sass/pages/_blog.scss
shredtechular: Great point.  I have removed the _blog.scss file entirely and created a separate components/_tablse.scss file.  In doing so, this removed the ``blog-page`` from the selector, so I think this will be more logical?

Addressed comment in https://github.com/m-lab/m-lab.github.io/pull/91/commits/cfd1724768b3c1f78fd772dc279e3cde618b3d6b

PR #107, opened 2016-03-17
Invalidate CloudFront cache after building a new, deployable version of site.

nkinkade:
When a new version of the website is built and deployed, we need to invalidate the CloudFront cache so that stale, cached content isn't being served up to users.  This change set requires [a new gem](https://rubygems.org/gems/cf-s3-invalidator/) to work.  Also included in this change set in a new file, which was encrypted with _travis encrypt-file_.  The encrypted file contains credentials for an Amazon IAM account which only has permissions to PUT into the S3 bucket and create cache invalidations.. the same account used for deployment.

The reason that the commands added to the .travis.yml file are wrapped in conditions is 1) we don't want to invalidate the cache on PR builds or non-master activity 2) PRs from forks don't have access to the secure environment variables needed to decrypt the AWS account credentials file.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/107)


.travis.yml
```diff
@@ -17,3 +17,12 @@ deploy:
   on:
     repo: m-lab/m-lab.github.io
     branch: master
+before_install:
+  if [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
+    openssl aes-256-cbc -K $encrypted_69b240b11ca3_key -iv $encrypted_69b240b11ca3_iv -in _cf_s3_invalidator.yml.enc -out _cf_s3_invalidator.yml -d;
```
mtlynch: Can we add documentation for what this is doing and why it's necessary? I think the documentation you wrote in the PR description is good. I'd like it to live with the code, so can we adapt it and bring it in here?

Similarly to the S3 key, can we include instructions in the README for how to regenerate this data if we need to roll creds?

mtlynch:
LGTM

.travis.yml
nkinkade: Documentation added to both the Travis config and the README.  Merging and crossing fingers that it works as is, but could needs some tweaking... we'll see.

Issue #104, opened 2016-03-17
De-minify desktop.min.js


mtlynch:
This file is minified:

https://github.com/m-lab/m-lab.github.io/blob/master/js/desktop.min.js

It's unclear if we have the un-minified source anywhere, but minified JS should not be in source control. We should de-minifiy this using a JS formatter that can apply the formatting rules of the [Google JS style guide](https://google.github.io/styleguide/javascriptguide.xml).

shredtechular:
I'm not able to close issues, so leaving a comment that I think PR #109 addressed this issue and I think it can be closed?

PR #109, opened 2016-03-19
De-minify javascript

shredtechular:
This PR fixes issue #104 which adds in source js files for the site.  This PR also is a pre-requisite for additional work on PR #83 .  Once the js de-minification & refactor is complete, then more clean-up work can happen on that PR in implementing improvements around anchor links.

The [original source js files](https://github.com/m-lab/website/tree/master/static/js/src) were used as a starting point for de-minifying the original desktop.min.js file.  There was a significant amount of unused js in those files, so only included what was absolutely necessary and tried to go through them and clean up where I thought it was needed.  Notes below:

- [x] De-minified desktop.min.js and removed it from version control.  During the js clean up was also able to remove 2 js libs that are not needed from the codebase (hoverintent & jquery migrate).

- [x] The other big piece of de-minifying the js was to ensure a method to compile and optimize the de-minified js back into one minified js file to be used in production environments.  The source js is using require.js to load the necessary js files & libraries, so I included the [r.js](http://requirejs.org/docs/optimization.html) library and a build profile (app.build.js) for doing the optimization of the js.
    - The travis-checks file was then modified to incorporate the commands for building the source js as well as incorporated an sed function to modify the _config-deploy build configuration for production so when building for production, the concatenated & minified js file is used.  As you can see on my [fork](http://shredtechular.github.io/m-lab.github.io/), and all forked repos, the un-minified js is being used.  The optimized js will only be used on production.
    - The compiled app.js file is also now included in the `.gitignore` to ensure it is not checked back into version control.

Additional notes for consideration of future tasks (considered out of scope for this PR, but if folks agree, we can work them in as well).
- [ ]  There are some older versions of js libraries being used.  It might be good to consider updating these to the latest versions to stay current.

- [ ]  There were a number of console.logs in the js code.  I didn't see anything about removing these in the [js style guide](https://google.github.io/styleguide/javascriptguide.xml) however, it might be good to remove these unless someone feels differently?  They are currently in the desktop.min.js so kept them in for now, but doesn't seem like they are needed for anything and looks like normal troubleshooting code?  Alternatively, we can look into automatically removing these statements as part of build scripts for production.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/109)


critzo:
:lgtm:

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/109#-:-KDFEgYk8d24f4q1eGXG:1417088721)*

PR #110, opened 2016-03-22
Re-doing "quick link" anchors

shredtechular:
## This PR supersedes pull #83 for adding anchor links to the data docs pages.  If we like this implementation then #83 can be closed as well as issue #77 .

As the lighter js has now been implemented into the site, this PR re-addresses the anchor link or table of content (TOC) capability on individual pages.  You will notice that an author no longer has to put in the `{:.anchor-classname}` syntax for each heading that should have an anchor.  An author does still need to specify each anchor in the yml frontmatter of each post however, and they **must match** the actual heading.  

If we want to avoid that yml frontmatter with these links, then we could implement either a jekyll plugin (won't work on gh-pages) or simple js that identifies headers on a page and dynamically inserts the TOC at the top of a page.  It would be possible to customize these to some degree (meaning show a TOC or not on each page as well as configure what headings should appear in a TOC sitewide) but we would need to define the business rule for that.  All headings in the TOC?  Only h2s?  h1s through h6s?

In the pages below, you will see that I did TOC links for H1s and H2s, but again, one advantage with the current implementation is it is completely customizable for each page.  The downside is additional manual work for an author.

Other notes to point out with this PR:
- You will see that with this implementation that the anchor links behave more intuitively and include the hash fragment in the URL when an anchor link is clicked.
- Needed to refactor the TOC logic in page.html slightly as there were headings with a dash in them which was causing issues with how kramdown is generating the heading ids.
- Some of the anchor links were quite long, so refactored the styling a tad to allow less wrapping of these links and attempting to utilize the space a little better.

These updates can be seen on my [fork](http://shredtechular.github.io/m-lab.github.io/).

### DIFF
--- a/_layouts/page.html
+++ b/_layouts/page.html
@@ -12,7 +12,7 @@ layout: default
           	
   	{% for grid in page.grid-section %}
 
-		<li><a href="#{{ grid.group-heading | downcase | replace: " ", "-" }}" data-proofer-ignore><i class="icon icon-triangle icon-right icon-before"></i>{{ grid.group-heading }}</a></li>
+		<li><a href="#{{ grid.group-heading | downcase | replace: " — ", "--" | replace: " ", "-" }}" data-proofer-ignore><i class="icon icon-triangle icon-right icon-before"></i>{{ grid.group-heading }}</a></li> 
 
 		{% endfor %}
 
@@ -25,7 +25,7 @@ layout: default
           	
   	{% for link in col.column %}
 
-		<li><a href="#{{ link.group-heading | downcase | replace: " ", "-" }}" data-proofer-ignore><i class="icon icon-triangle icon-right icon-before"></i>{{ link.group-heading }}</a></li>
+		<li><a href="#{{ link.group-heading | downcase | replace: " — ", "--" | replace: " ", "-" }}" data-proofer-ignore><i class="icon icon-triangle icon-right icon-before"></i>{{ link.group-heading }}</a></li>
 
 		{% endfor %}
 
--- a/_pages/bigquery_examples.md
+++ b/_pages/bigquery_examples.md
@@ -3,6 +3,16 @@ layout: page
 title: "BigQuery Examples"
 permalink: /data/bq/examples/
 breadcrumb: data
+accordion-quick-links: true
+quick-links-section:
+  - column:
+    - group-heading: "Overview"
+    - group-heading: "Basic counting"
+    - group-heading: "Computing statistics over time"
+  - column:
+    - group-heading: "Dealing with IP addresses"
+    - group-heading: "Comparing NDT and NPAD tests"
+    - group-heading: "Computing distributions of tests across users"
 ---
 
 # BigQuery Examples
--- a/_pages/bigquery_examples.md
+++ b/_pages/bigquery_examples.md
@@ -6,13 +6,13 @@ breadcrumb: data
 accordion-quick-links: true
 quick-links-section:
   - column:
-    - group-heading: "Overview"
-    - group-heading: "Basic counting"
-    - group-heading: "Computing statistics over time"
+    - group-heading: "BigQuery Examples"
+    - group-heading: "Basic counting — How many users"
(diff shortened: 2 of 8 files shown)


critzo:
I think that even though there's a bit of extra work for authors, the ability to customize each page would be ideal. 

Commenting on the features implemented in this PR:

- I would rather be required to add {:.anchor-classname} where I wanted anchors than have them be super long, matching the text of the heading itself. 
- I don't like the idea of creating a business rule for auto-generating anchors. 
- I also would like the ability to add front-matter to indicate whether a TOC at the top of the page should be shown or not

@mtlynch @georgiabullen what do you think?

mtlynch:
Ideal for me is to just have h2 & h3 elements (maybe all h's?) automatically allow the user to create anchor links to them by either clicking the heading or clicking an icon near the heading.

I don't think I should really have a vote in this because I won't have much bandwidth in the next few weeks to stay closely engaged in this decision, but those are my 2 cents.

critzo:
Yeah, I'd like to hear what @georgiamoon thinks. In theory an id=anchorName
element can be included in any html tag, not just H elements, but I think
we'll just be using them for H elements. I think we want to customize which
ones get them, and the ability to show or not show the TOC links at the top
of the page. I was also wondering what a left or right side menu would look
like for the TOC, especially on the FAQ page, which is going to have a lot.

critzo:
@mtlynch and @shredtechular 
@georgiamoon and I spoke today and have decided the following about the anchors:
* First a question: can we have anchors automatically generated for all H tags **and** the ability to decide which would be listed in the TOC **and** also have a flag in the frontmatter that determines whether the TOC is displayed?
* If that's a big lift or not possible, we want to go with the following:
  * authors will manually add anchors where they want them and add frontmatter when they want a TOC.
  * while initially we will likely use anchors only on H1-H6 tags, they should be able to be applied to any element

shredtechular:
@critzo It is absolutely possible to automatically generate anchors for all H tags with a little js **and** still keep the existing flag in the frontmatter that determines whether the TOC is displayed, though we may want to change the title of it to make more sense.  (currently, the `accordion-quick-links: true` will display the TOC and if it does not exist or is false, then it will **not** show.).

Thank you for the feedback & decision on that!  One last question, you mentioned you were wondering how a side menu for the TOC links might look?  One thing I thought of when you mentioned that was a spyscroll type of functionality so when a user is scrolling down, they could see where they were on the page as they scrolled down.  Or they could simply click on of the TOC/jump links to go straight to that section.  Something similar to how the TOC on the right is shown on the bootstrap site maybe?  For example: [http://getbootstrap.com/javascript/#scrollspy](http://getbootstrap.com/javascript/#scrollspy)  That might be getting more complex that what you want/need, but thought I'd through it out there since it was brought up...

critzo:
@shredtechular I haven't thought much yet about how the side menu might look or function, but I do like the example in the scrollspy repo. We should get more thoughts on this from @georgiamoon . I was mostly thinking that depending on the number of anchors, a TOC could grow rather larger and use up a bunch of screen real estate at the top of the page. The other concern I had was with the length of the auto generated anchor names. Lastly, and this is a new comment, if an anchor is clicked in the TOC or if opened in the browser (link from another page for example) can the accordion fornthat element be opened automatically as well?

georgiamoon:
scrollspy looks cool -- seems like a nice to have.

shredtechular:
@critzo @georgiamoon Understood on the scrollspy being a nice to have and I think that would be something we could look at once we implement bootstrap into the site.  I do share some of the same concerns about the TOC taking up more real estate than expected if it auto generates for all H tags, but we can see how it goes and going with the theme that is on the /faqs/ page, I think we can break them into 2 columns if there is more than say 4 auto generated TOC elements.

@critzo that is a good point about opening an accordion header if a user clicks on an applicable H tag that is an accordion.  I'm pretty sure that should be doable and will try to incorporate that functionality as well in the next PR.

I think I'm all set on a direction to head with this now, I'll get to work on this!

critzo:
This PR was superceded by PR #119 
Closing.

PR #113, opened 2016-03-24
Fix old or broken links on NDT test page.

nkinkade:
* Links to the NDT source were still pointing to code.google.com, which is defunct.
* There was a link to "Download NDT client", but no such thing exists, and least not on the Internet2 site or Github, and the links simply pointed to the source, which was misleading.... so remove that link altogether.
* The "more information" bullet was pointing redundantly to the source, and also pointing to a page at internet2.org which no longer existed.
* The canonical M-Lab website address is prefixed with www, so all links should point to that domain.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/113)


critzo:
@nkinkade LGTM!

PR #114, opened 2016-03-24
Fix breaking openssl decrypt and invalidate CF cache after deploy.

nkinkade:
PR #107 (merged) attempted to set up CloudFront cache invalidation after a build.  After the PR was already merged, @mtlynch discovered a few optimizations that needed to be made:

* Can we add _cf_s3_invalidator.yml to the .gitignore to minimize the chance of someone ever accidentally checking this in unencrypted?
* Does it hurt anything if we excessively flush the CloudFront cache a few times? If not, I suggest in the next PR to just remove the [ "$TRAVIS_PULL_REQUEST" == "false" ] conditions so that we don't have to actually merge to figure out if it's working, then we can put that condition back in once we get it working.
* also suggest using after_deploy instead of after_success so that our builds break if cache validation is failing. 

Additionally, I noticed that the merge of #107 caused the Travis build to fail at the point where Travis is attempting to decrypt the AWS credentials in _cf_s3_invalidate.yml.enc. I reencrypted the file using, I think, the same keys for the m-lab/mlab.github.io repo, and yet the resultant encrypted file differed, so maybe I did something wrong in encrypting the file the first time.  This PR also includes the updated encrypted file.

Regarding point 2 above from @mtlynch, we could remove the  [ "$TRAVIS_PULL_REQUEST" == "false" ] from the _after_deploy_ block, but not from the _before_install_ block because the necessary secret environment variables to decrypt the file are not available to PRs from forks for the repo.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/114)


mtlynch:
:lgtm:

---

Reviewed 3 of 3 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/114#-:-KDdO5FmAesc8Tr1Z-05:1417088721)*

PR #115, opened 2016-03-24
Force Travis build to fail when invalidating the CF cache fails.

nkinkade:
The cf-s3-inv command failing after the deploy does not cause Travis to fail the whole build, yet we want any sort of failure to fail the whole thing, including if CloudFront cache invalidation fails.  This add an _|| travis_terminate 1_ after the _cf-s3-inv_ command with the idea that if cf-s3-inv fails it will force the entire Travis build to exit with a failing status (1).

An additional note is that I think I've corrected the reason for the failure in AWS.  Apparently the AWS user didn't have sufficient privileges...it needed cloudfront:GetDistribution permission (according to the error text), which the user now has.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/115)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,5 +29,5 @@ before_install:
 after_success:
   # Only invalidate CloudFront cache if this on the master branch and is not a PR
   #if [ "$TRAVIS_BRANCH" == "master" ] && [ "$TRAVIS_PULL_REQUEST" == "false" ]; then
-  cf-s3-inv || travis_terminate 1;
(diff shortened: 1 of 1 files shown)


mtlynch:
LGTM

>Apparently the AWS user didn't have sufficient privileges...it needed cloudfront:GetDistribution permission (according to the error text), which the user now has.

Can you actually change this back temporarily so that we can verify this will fail loudly? Afterwards, I think we can force a Travis re-run from the Travis dashboard.

PR #119, opened 2016-03-28
Auto-generation of Table of Contents

shredtechular:
## This PR supersedes pull #83 and #110 for adding anchor links to the data docs pages.  If we like this implementation then these can be closed as well as issue #77 .

As the lighter js has now been implemented into the site, this PR re-addresses the anchor link or table of content (TOC) capability on individual pages.  You will notice that an author no longer has to put in the `{:.anchor-classname}` syntax for each heading that should have an anchor.  

You will also notice that an author does **NOT** need to specify each anchor in the yml frontmatter of each post either.

With this PR, if an author would like to display a Table Of Contents anywhere within their post, they can simply add the following to the markdown file:
```
* Table of Contents
{:toc}
```  

If an author needs more control over the customizing of the ToC, then they can use the existing format (as in the /tests/ page still using the old customized ToC).  For example, if it is desired to not show ToC links for certain h tags, this method can still be used for customization purposes.

If neither of those exist, then no ToC will be shown.

Other notes to point out with this PR:
- The auto-generated ToC will not split into 2 columns, this PR only supports 1 column ToC's.
- The auto-generated ToC does create indentation/nesting levels automatically as well depending on the H tag number.
- You will see that with this implementation that the anchor links behave more intuitively and include the hash fragment in the URL when an anchor link is clicked.
- Some of the anchor links were quite long, so refactored the styling a tad to allow less wrapping of these links and attempting to utilize the space a little better.

These updates can be seen on my [fork](http://shredtechular.github.io/m-lab.github.io/).

**To ensure the scope of this PR is understood, the following are planned for future PRs:**
- [ ] When clicking an accordion ToC link, the accordion automatically opens.
- [ ] GFM link icons to appear on left of H tags to let user know they can bookmark/go straight to that link
- [ ] Splitting into two columns if more than 4 and or scrollspy if desired in the future (dependency on bootstrap)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -19,13 +19,18 @@ permalink: "/blog/:title/"
 sass:
   sass_dir: _sass
   style: compressed
-markdown: kramdown
 include: ['_pages']
 exclude:
   - .gitignore
   - README.md
   - Gemfile
   - Gemfile.lock
+
+# Markdown settings
+markdown: kramdown
+kramdown:
+  auto_ids:       true
+  auto_id_stripping:    true
 
 # Display blog posts even if they're future-dated.
 future: true
--- a/_pages/bigquery_examples.md
+++ b/_pages/bigquery_examples.md
@@ -5,6 +5,9 @@ permalink: /data/bq/examples/
 breadcrumb: data
 ---
 
+* Table of Contents
+{:toc}
+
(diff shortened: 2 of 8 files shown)


critzo:
LGTM. Merging.

PR #133, opened 2016-03-30
Removing Accordions

shredtechular:
The focus of this PR is to remove the accordions from the site, which was the brainstormed solution for ensuring anchors were put on all areas of the site that need them. This solution optimizes the ease of editing markdown files and keeps all content within Markdown as opposed to over-engineered clutter to keep track of in front-matter key/values & code snippets.  This solution keeps things nice and simple and offers ways for advanced content editors to customize anchor names as well as the ability to keep certain headers from being autogenerated into the TOC jump links.

You can preview the accordion removal on my [fork](http://shredtechular.github.io/m-lab.github.io/faq/).  The key things to observe is that all h2s that used to be in accordions now have ids that can be linked too and how the page looks/works without accordions to be sure this is an acceptable solution.

Additional details of this PR:
- [X] The accordion include is now completely removed from the site.

- [X] All markdown files that were currently using accordion functionality have been updated to move all the "header" frontmatter values into the markdown as well as cleanup remaining unneeded frontmatter keys & code snippets that were only serving accordion functionality purposes.

- [X] If an author determines that a header should *NOT* be put into a ToC, then a user can simply add `{: .no-toc }` underneath a header.  This was necessary on the /faq/ page to prevent an unwieldly ToC on that page until we can move that to a sidebar.

- [X] If an author would like to override the anchor values, then they may simply add `{: #preferred-anchor-name }` above a header.  This example would give the header tag an id of "preferred-anchor-name".  

- [X] The last thing to note is from removing the accordions, the markdown linter popped up a new issue on the publications page about having the same content in 2 different headers on the same markdown file.  For example, the header "2012" is in 2 places on the /publications/ page.  The linter documenation states that the rationale behind this rule is that, "Some markdown parses generate anchors for headers based on the header name, and having headers with the same content can cause problems with this."  In the case of Kramdown markdown parsing however, it will automatically append a "-XX" to headers that have the same exact content/ids and increment the XX by 1 for each occurance of the same id.  Since Kramdown handles this situation, I think we are okay to add this rule to the exclusions list.

- [ ] Next PR will address adding some js to the site so a link icon will appear next to anchors upon hovering over that item.

Future considerations:
- May want to review styling of the FAQs page and ensuring users can clearly see beginning/ends to FAQ sections as we want to make these as simple as possible for users to read through.

- May want to consider `<dl>` for content such as FAQs.  It seems like this would be more semantic for this type of content, although we did run into issues generating ids for `<dt>`s.  Not something immediate, but still thought worthwhile to mention as there are benefits to semantic html as much as possible.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/133)


critzo:
:lgtm:

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/133#-:-KEBp2rpRPaIewEZMHHP:1417088721)*

critzo:
Merging.

PR #134, opened 2016-03-30
gitignore Update

shredtechular:
This PR is a very simple update to add another file that should never be committed to git and I know I've almost done it a few times by accident.  This file is generated when the sed commands are editing the _config file on the fly (like a temp file, so it shouldn't ever be checked in).

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/134)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,5 +14,6 @@ coverage
 .ruby-version
 .sass-cache
 _config-deploy.yml
+_config-deploy.ymle
 js/app.js
 _cf_s3_invalidator.yml


critzo:
:lgtm:

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/134#-:-KEBqwnXprNvf0RFj7Y5:1417088721)*

critzo:
@shredtechular LGTM. Merging.

Issue #138, opened 2016-03-31
Strange bullets in front of Observatory Menu buttons


georgiamoon:
See here:
![screen shot 2016-03-31 at 3 28 01 pm](https://cloud.githubusercontent.com/assets/941210/14188274/4fa05662-f755-11e5-939d-6acc82230ff9.png)

and here:
![screen shot 2016-03-31 at 3 28 19 pm](https://cloud.githubusercontent.com/assets/941210/14188283/54e2ef2c-f755-11e5-803e-6ce9978b18be.png)

I *think* that these are just css issues with what is probably a `<ul>` element.

critzo:
This issue should be resolved by #139. Waiting for cache to refresh to close.

PR #139, opened 2016-03-31
Bug - Observatory List Items

shredtechular:
Saw a bug come in that @georgiamoon caught and documented in issue #138.  Excellent catch BTW!

I think there are some other cases of conflicting styles happening between codebases on that page (i.e. - the main nav, no footer, etc), so we can probably tighten this up even more as well.  But this is one item that I know I introduced with the SASS styling cutover, so wanted to get this fix in at least since I know exactly what was causing that.

This fix changes the class to an id & renames it so we shouldn't have to worry about any specific classes for native pages overriding the submodules styling.  The only issue now would be base classes from each codebase, so needed to target the list items within that new section id to get it back to the intended submodule styling.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/139)

### DIFF
--- a/_layouts/observatory.html
+++ b/_layouts/observatory.html
@@ -3,7 +3,7 @@ layout: default
 ---
 
 <div class="subpage-content full-width observatory-page">
-	<section class="intro-section">
+	<section id="intro-section-observatory">
 		<h1 class="section-heading">{{ page.heading }}</h1>
 
 		<div class="wrapper">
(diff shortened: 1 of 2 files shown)


critzo:
:lgtm:

---







*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/139#-:-KEDQvtkFx6_B49vaCPE:1417088721)*

PR #145, opened 2016-04-05
Feature Responsive: Incorporation of Bootstrap JS

shredtechular:
This is a quick PR that incorporates the compiled bootstrap js library into the MLab through Require.js.  

Since we are utilizing Require.js, it also became apparent to me that we needed to move the bootstrap js library into what requirejs could access and use, not to mention to have all libs in one place.

I think a future task could be to do a little directory restructuring after the responsive feature that moves images, fonts, SASS/CSS, and js into a consolidated "assets" directory.  This would be more in line with how bootstrap organizes their files and I think there are some advantages to doing that.

Until that dir re-structure happens though, I think its best to have the bootstrap library in /js/libs/ directory for now.


shredtechular:
Closing, made against wrong branch.

PR #142, opened 2016-04-01
Feature Responsive: Incorporation of Bootstrap - Part 2

shredtechular:
~~~#### Please NOTE: This PR is dependent on #141.  This PR should not be merged in unless #141 is accepted and merged.~~~

~~~#### Additionally, this PR will fail until #141 is merged and this PR can be rebased so it can find bootstrap references.~~~ 
**Update:** PR #141 was merged & this PR has now been rebased.

This PR is the second part of implementing bootstrap into the site with an end goal of making the site fully responsive.  This PR was made to show that 9 files needed to be amended in order to accompany bootstrap into the site as there were several styles that bootstrap was overriding.  The changes you see here, make it so that when this PR is accepted/merged, there are no style changes to the existing site with the incorporation of bootstrap.

Once this PR is accpeted/merged, then future pull requests will then have visual progress towards making the site fully responsive.

Additional PR notes:
- To begin with I have set bootstrap up to compile into its own and separate css file.  Upon conclusion of completing the responsive behavior throughout the site, then we can investigate combining bootstrap.css & base.css into 1 css file that the site will need to load as opposed to both.  For now, I'd advise on keeping separate until the responsive behavior has been completed in full.

- While fixing styles that bootstrap was overriding, found it most efficient to start to port those styles over into our SASS structure as well.  So, was able to refactor 3 components (buttons, links, and carousel) and remove those styles from the original CSS.

<!-- Reviewable:start -->
---
This change is [<img src="https://reviewable.io/review_button.svg" height="35" align="absmiddle" alt="Reviewable"/>](https://reviewable.io/reviews/m-lab/m-lab.github.io/142)
<!-- Reviewable:end -->

PR #141, opened 2016-04-01
Feature Responsive: Incorporation of Bootstrap - Part 1

shredtechular:
This PR focuses on the straight implementation of bootstrap into the site.  It is a rather large amount of files, so Part 2 will be focused on the modifications of current styles to ensure no style changes happen when incorporating bootstrap into the site and finally adding the compiled bootstrap css file into the site.  This is the first step towards making the site fully responsive.

This is obviously a large amount of files to introduce into the site, however, there isn't any code changes here, this is a straight download of bootstrap v3.3.6 and all this is doing is adding these files into the site.  **Nothing in this PR will make bootstrap "active" by referencing styles of javascript on the site.**

There were 72 source SASS files that come with bootstrap v3.3.6, and 20 js & font files.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/141)

PR #148, opened 2016-04-05
Feature Responsive: Consolidation & Customization of Bootstrap Styling

shredtechular:
PR's #141 & #142 pulled in the Bootstrap files & then fixed styles that Bootstrap was overriding.

This PR continues a little more Bootstrap implementation & customization work that you can see listed below:

1. The first thing to point out is the new `_sass/helpers/_custom-bootstrap-variables.scss` file.  This file is a straight copy of the default bootstrap variables file, but it is recommended to make a copy of it and store it within our custom SASS files if we needed to customize any bootstrap variables.  Due to our directory structure not matching completely what Bootstrap was expecting, we did need to customize one path variable (`icon-font-path` on line 83) for the iconography/glyphicon fonts.  This file gives us the ability to easily customize any other default bootstrap variables as well in the future.

2. Since we needed to customize a bootstrap variable, we need to go ahead and consolidate bootstrap.css & base.css into one CSS file that the site can use.  You can see this work in `_includes/head.html`, `css/base.scss`.

3. Finally a small misc cleanup item that removes some legacy straight CSS and puts them into our base.scss file as we continue to move all legacy CSS into SASS.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/148)

### DIFF
--- a/_includes/head.html
+++ b/_includes/head.html
@@ -31,7 +31,6 @@
   {% feed_meta %}
 
   <!-- Styles -->
-  <link href="{{ site.baseurl }}/css/bootstrap.css" rel="stylesheet" />
   <link href="{{ site.baseurl }}/css/base.css" rel="stylesheet" />
 
   {% if page.audience == "observatory" %}
--- a/_sass/base/_base.scss
+++ b/_sass/base/_base.scss
@@ -3,6 +3,18 @@
 //import mixins
 @import 'helpers/_mixins';
 
+* {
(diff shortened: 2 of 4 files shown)


critzo:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/148#-:-KEbEBy0Zvy3JeCEeQEH:1417088721)*

PR #153, opened 2016-04-05
Feature Responsive: Layout & style updates for responsive footer

shredtechular:
This PR adds layout and style updates for making the footer a bit more responsive.  

These changes can be previewed on my [fork](https://shredtechular.github.io/m-lab.github.io/).  You can now slowly change the width of your browser and all portions of the footer breakdown quite nicely.

Below are a few screen caps of how the responsiveness of the footer is looking at a few different breakpoints:

### Desktop:
screen shot 2016-04-05 at 6 29 02 pm

### Tablet:
screen shot 2016-04-05 at 6 29 27 pm

#### Small Mobile/Phone:
screen shot 2016-04-05 at 6 29 45 pm

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/153)

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -13,46 +13,53 @@
 <footer class="footer-section">
   <span class="footer-shadow"></span>
 
-  <!-- Footer - Social Media -->
-  <div class="footer-ctn">
-    <div class="callout-col l-span1">
-      <h2>{{ footer.subscribe.title }}</h2>
-      <nav>
-        <ul class="callout-ctn social-ctn">
+  <div class="container">
+    <div class="row">
 
-			  {% for item in footer.subscribe.items %}
-          <li><a href="{% if item.link | slice: 0 = "/" %}{{ item.link | prepend: site.baseurl }}{% else %}{{ item.link }}{% endif %}" class="btn-social {{ item.link-class }}" target="_blank">{{ item.link-title }}</a>
-          </li>
-			  {% endfor %}
+        <!-- Footer - Social Media -->
+        <div class="col-xs-7 col-sm-5 col-md-4">
+          <h2>{{ footer.subscribe.title }}</h2>
+          <nav>
+            <ul class="callout-ctn social-ctn">
 
-        </ul>
-      </nav>
+    			  {% for item in footer.subscribe.items %}
+              <li><a href="{% if item.link | slice: 0 = "/" %}{{ item.link | prepend: site.baseurl }}{% else %}{{ item.link }}{% endif %}" class="btn-social {{ item.link-class }}" target="_blank">{{ item.link-title }}</a>
+              </li>
+    			  {% endfor %}
+
+            </ul>
+          </nav>
+        </div>
+      <!-- Footer - Navigational Links -->
+        <div class="col-sm-6 col-md-6 pull-left">
+          <ul class="footer-links">
+    			  
+    			  {% for item in footer.other-links.column-one.items %}
+              <li><a href="{{ item.link | prepend: site.baseurl }}">{{ item.link-title }}</a>
+              </li>
+    			  {% endfor %}
+
+          </ul>
+          <ul class="footer-links">
+            
+    			  {% for item in footer.other-links.column-two.items %}
+              <li><a href="{{ item.link | prepend: site.baseurl }}">{{ item.link-title }}</a>
+              </li>
+    			  {% endfor %}
+          
+          </ul>
+        </div>
+
     </div>
+    <div class="row">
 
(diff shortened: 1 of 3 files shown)


critzo:
For the most part, :lgtm:  -- One comment to flag for further PRs on the footer responsiveness-- testing this on a desktop browser (Chrome), when viewing at the mobile width, the "back to top" button covers the privacy policy text.  I think this can be dealt with as you begin to work on the responsiveness of content in general, and maybe in a final browser and device check before the responsiveness is pushed to master.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/153#-:-KEfVehPYwc7cD-2FCts:1384765752)*

shredtechular:
Good catch @critzo .  I forgot to push one commit I did to make it so the back to top button didn't cover the privacy policy text to my forked gh-pages branch.  I did include it with this pull request but failed to push that same fix to my gh-pages branch.  I have pushed that tweak to my gh-pages branch now if you wanted to retest that and can see it won't cover the footer items since we have space there.

I think this does bring up a good point though and we will need to see what we can do with that back to top button from a general content perspective.  I will see what solutions we can come up with to gracefully handle that button which will probably include removing the "back to top" text to make that button smaller on mobile devices...but will see what we can do there in an upcoming PR.

georgiamoon:
Devil's advocate question: I would argue that the "back to top" button as
implemented is not a great design/UX (maybe a bit of an anti-pattern even?).

what if we:
1) removed it entirely? OR
2) Changed it to a link that *only* appears on long pages where one might
reasonably need a "back to top" button (e.g. the FAQ, more like in a long
digest email)

critzo:
For longer pages on mobile, I definitely used it, but maybe it's less useful on other pages.

shredtechular:
1) Removing it entirely is obviously really easy to do obviously!  But I do think it could serve a useful purpose especially on mobile devices as the pages tend to get longer there...  It looks like the current logic is if a user scrolls down more than 500 pixels that is when the button would show.  If we think it is showing up too soon we can change that setting.

2) This makes sense to me as well.  We could add a static back to top link to headers (right side?) and then in the footer too if I'm understanding your intent correctly there.  Though it might be unnecessarily complicated to only do this on longer pages?

3) Another idea I'll throw out there would be a sticky header that I think some sites are using now as a different way to solve the "back to top" issue.  In the current design of the desktop header, I'm not sure it lends itself to converting to a sticky header very nicely, but on mobile I think it might work well.  Maybe we could finagle or update the desktop header a bit so a sticky header might work well for all browsers.  Just a thought I'd throw out there that I think also attempts to solve the same type of navigational issues.

From my perspective, and in the short term, the important thing *right now* with the back-to-top button is making sure it doesn't overlap footer content.  I think it is okay if it overlaps body text content?  Since the user can just keep scrolling to see the text?  But if it was overlapping footer content then I think that would be a problem since the user would be out of scrolling room.  Does that make sense?

georgiamoon:
>
> 1) Removing it entirely is obviously really easy to do obviously! But I do
> think it could serve a useful purpose especially on mobile devices as the
> pages tend to get longer there... It looks like the current logic is if a
> user scrolls down more than 500 pixels that is when the button would show.
> If we think it is showing up too soon we can change that setting.
>
2) This makes sense to me as well. We could add a static back to top link
> to headers (right side?) and then in the footer too if I'm understanding
> your intent correctly there. Though it might be unnecessarily complicated
> to only do this on longer pages?
>
On all pages is fine, and yes, my suggestion is in headers on the right
side, and a link at the footer.

> 3) Another idea I'll throw out there would be a sticky header that I think
> some sites are using now as a different way to solve the "back to top"
> issue. In the current design of the desktop header, I'm not sure it lends
> itself to converting to a sticky header very nicely, but on mobile I think
> it might work well. Maybe we could finagle or update the desktop header a
> bit so a sticky header might work well for all browsers. Just a thought I'd
> throw out there that I think also attempts to solve the same type of
> navigational issues.
>
Agreed -- this is a common pattern that could address this. Let's hold off
for now, and look at that after we're done with mobile responsiveness.

> From my perspective, and in the short term, the important thing *right
> now* with the back-to-top button is making sure it doesn't overlap footer
> content. I think it is okay if it overlaps body text content? Since the
> user can just keep scrolling to see the text? But if it was overlapping
> footer content then I think that would be a problem since the user would be
> out of scrolling room. Does that make sense?
>
I think for now that's fine, and let's fix it in the ways described above
by moving to a link rather than the button ASAP. I'd prefer to not have it
blocking anything.

PR #155, opened 2016-04-06
Feature Responsive: Update for anchors

shredtechular:
This is a very minor update that will make anchors a little more responsive on mobile browser widths.  Users on mobile devices probably won't even see these, but folks that re-size their browsers may notice, so this style change makes the anchors fit a little more nicely on very small browser widths.

This change can be previewed on my [fork](https://shredtechular.github.io/m-lab.github.io/).

### Before:
anchors_bad

### After:
anchors-good

### DIFF
--- a/_sass/components/_anchors.scss
+++ b/_sass/components/_anchors.scss
@@ -11,6 +11,11 @@
   margin-left: -20px;
   font-size: 20px;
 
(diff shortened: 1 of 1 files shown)


critzo:
@shredtechular LGTM! Merging.

PR #157, opened 2016-04-07
Feature Responsive: Homepage

shredtechular:
This PR adds layout and style updates for making the rest of the homepage responsive.  

These changes can be previewed on my [fork](https://shredtechular.github.io/m-lab.github.io/).  You can now slowly change the width of your browser and all portions of the homepage are responsive and adjust to changing widths.

Notes:
- One area I think is a little odd is the latest blog posts.  Something that I think may utilize the space a little better is to stack the 2 latest blog posts even on desktop.  This way we could use a consistent font size on them no matter the length and I think it looks a little better, especially as the width breaks down.  I don't see that as a showstopper for this pull request, but a design choice that could be improved upon if folks would like to go that route.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/157)

### DIFF
--- a/_layouts/homepage.html
+++ b/_layouts/homepage.html
@@ -7,77 +7,81 @@ layout: default
 <meta data-video-id="{{ page.videoid }}">
 
 <div class="about-section info-segment">
-  <div class="info-segment-inner">
-    <section class="what-is-col l-span2">
-        <h2 class="info-heading initial-heading">{{ page.info-heading }}</h2>
-        <article class="video-entry">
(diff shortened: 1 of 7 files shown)


critzo:
:lgtm: With regard to your suggestion about stacking the two latest blog posts in that 2/3 width div at desktop widths makes sense to me. Let's do that in a future PR. 

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/157#-:-KEkctcQDvox0QvVXDsl:238040505)*

PR #158, opened 2016-04-07
Feature Responsive: Blogs

shredtechular:
This PR adds layout and style updates for making the blog layouts (index, archive, and single blog pages) responsive.  

These changes can be previewed on my [fork](https://shredtechular.github.io/m-lab.github.io/) and you can test the responsive nature of all the blog layouts.

Notes:
- Probably the biggest design decision with this PR is what to do with the blog sidebar when it breaks down to very small viewing widths.  In my opinion, the best option, short term, is to have the sidebar collapse underneath the content.  The content is more important than the sidebar and I think it is okay that the sidebar displays underneath so a user can still use that as a navigational tool.  Future considerations might include more fancier solutions, some that I can think of below are:
    - Slide the sidebar out from the side when a button is clicked.  This might be the nicest solution, but probably a heavier LOE.
    - Maybe have a couple of dropdowns or accordions at the top for the 2 blog sidebar nav items?  Though,  I fear this could get a little confusing with the overall nav menu?
    - We could hide the sidebar altogether as well, if folks think its just not useful for mobile users?  Figured best to keep for now and just stack underneath.
- There were a couple of blog items that I tweaked the markdown on to get a couple of the images that are offset to the right to display a little nicer across all breakpoints.  They were really minor tweaks, but wanted to call that out as you will see a couple edits in .md files.
- [ ] One of the blogs has a table in it that isn't fully responsive yet.  This will come in a future PR.

Other than that, all the blogs are finished with this PR!

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/158)

### DIFF
--- a/_includes/siderail.html
+++ b/_includes/siderail.html
@@ -11,7 +11,7 @@
 {% endcomment %}
 
 <!-- Siderail -->
-<div class="blogroll-navigation l-span1 last">
+<div class="blogroll-navigation col-xs-12 col-sm-3 col-md-3 col-lg-2">
 
   <!-- Category Archives -->
   <div class="filter-col-ctn blog-cat-col">
--- a/_layouts/blog-archive.html
+++ b/_layouts/blog-archive.html
@@ -10,37 +10,48 @@ layout: default
   {% assign posts = site.posts %}
 {% endif %}
 
-{% include breadcrumbs.html %}
+<div class="container">
+  <div class="row">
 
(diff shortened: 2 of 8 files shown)


critzo:
:lgtm:

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/158#-:-KElRFlaCi87pPZYmTPi:1417088721)*

PR #159, opened 2016-04-07
Feature Responsive: Easy Fix for Responsive Data Tables

shredtechular:
This PR is for a low effort solution to make couple of the bigger data tables that exist on the site today a little more responsive.

As you can see in the code changes below, the fix is just a couple of lines of CSS.  I think this is a reasonable solution, but want to get a review on this.  The one downside that I see is it could result in a word breaking and going to the next line in bigger tables.  As you are probably aware, data tables are inherently not very responsive as the table columns have minimum widths that it must have before it just won't shrink anymore.  The CSS is a work around for that behavior.

If this is not acceptable, we can try some alternative solutions or introduce some 3rd party tools.  A couple of other options may be:
- [Bootstrap](http://getbootstrap.com/css/#tables-responsive) can add horizontal bars to tables so users can scroll left/right on tables, but does junk up the markdown a little more as an author would need to add a container `div` around tables.
- Another clever way to handle with CSS, which might be a little more LOE is outlined [here](https://css-tricks.com/responsive-data-tables/).  I have some concerns about complicating the authoring of tables, cross-browser compatibility and possibly accessibility with this solution, but we can give this a shot.
- Or there are some other third party tools that I can look into more to see if they offer other, better solutions.

These changes can be previewed on my [fork](https://shredtechular.github.io/m-lab.github.io/).

Below is a screen cap illustrating some of the breaking in the middle of a word on which smaller devices:
big-table-resp

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/159)

### DIFF
--- a/_sass/base/_base.scss
+++ b/_sass/base/_base.scss
@@ -30,6 +30,7 @@ caption,th,td {
   vertical-align: middle;
   word-wrap: break-word;
   max-width: 460px;
+  min-width: 75px;
 }
 
 tr {
--- a/_sass/components/_tables.scss
+++ b/_sass/components/_tables.scss
@@ -2,6 +2,8 @@
 
 
 // Customized bigquery table
-table#bigquery-schema-fields {
-  @include tables(200px);
+.bigquery.schema-page table {
+  @media screen and (max-width: $screen-md-min){
+    word-break: break-all;
+  }
 }


critzo:
@shredtechular I'm ok with this for now, but I don't really like the words breaking up on the end of lines, so I'd be in favor of the Bootstrap solution in the longer term. @georgiamoon thoughts on this?

PR #172, opened 2016-04-09
Finish conversion of css to SASS

shredtechular:
This PR moves the rest of the legacy CSS over to SASS preprocessor so it should be much more organized and manageable moving forward.

This is more for organizational purposes and no changes should be seen from a visual perspective.  This update is up on my [fork](http://shredtechular.github.io/m-lab.github.io/) to preview if desired.

Notes:
-  The majority of this PR was moving the legacy CSS into more organized SASS files.
-  There is still probably some opportunities to utilize variables and mixins to organize some of the SASS a little more, but this PR's goal is to move the remaining CSS over to SASS and remove selectors that are no longer in use.
-  Also, a couple of layouts and markdown files were updated where certain classes were not needed anymore with the bootstrap implementation.


critzo:
@shredtechular LGTM. Merging.

PR #173, opened 2016-04-09
Feature Responsive: Bootstrap Data Tables Capability

shredtechular:
Per comments in https://github.com/m-lab/m-lab.github.io/pull/159, this moves the solution for bigger responsive tables to bootstrap's responsive tables solution.

This removes the existing larger table's css to not break lines in the middle of a word and then in order to use bootstrap's solution, you can see it is enabled by wrapping a table in a `<div class="table-responsive" markdown="1"></div>` div.  The `markdown="1"` is critical to making the markdown'ed table render properly.  So, this will need to be enabled on a case-by-case basis as authors create/edit data tables on the site, so it does complicate the markdown a little.  But it will enable horizontal scrolling for large data tables.

I have put this up on my [fork](http://shredtechular.github.io/m-lab.github.io/) to be previewed if this solution is desired for large responsive data tables.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/173)

### DIFF
--- a/_pages/bigquery_schema.md
+++ b/_pages/bigquery_schema.md
@@ -42,6 +42,8 @@ M-Lab continues to support these tables, but urges clients to move to the new pe
 
 All M-Lab data shares the same data schema in BigQuery. The fields are described in the table below:
 
+<div class="table-responsive" markdown="1">
+
 | Field name                                           |     Type     |  Description                              |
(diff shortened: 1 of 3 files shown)


critzo:
@shredtechular LGTM. Merged.

PR #183, opened 2016-04-15
Feature Responsive: Making NDT test responsive

shredtechular:
This PR contains layout and style changes to make the NDT responsive within the iframe on the m-lab website.  

This update can be previewed on my [fork](http://shredtechular.github.io/m-lab.github.io/tools/ndt/).

Implementation details:
- One of the more challenging aspects of this is the usage of the iframe to incorporate the widget into the m-lab website as iframes don't automatically resize by nature (and without the use of 3rd party/addtl js the height still doesn't resize as content changes within the iframe).  To at least address the responsive widths for the iframe, I used bootstrap responsive iframe classes to wrap the iframe in the ndt markdown file.  I did slightly override the bootstrap classes to a more custom aspect ration of 54% and set a minimum height on the iframe so it doesn't go below a certain height cutting content if the device is resized to a small width.  So, this solution should prevent content from being cutoff while not having huge whitespace on desktop viewing sizes.  
- Also ensured scrollbars would appear on the iframe if content did grow outside the defined height.  For the most part this should not occur (tested in Chrome, FF, Safari, iphone 6, and several device emulators in Chrome), with the exception on the results summary screen.  The content on a couple of those tabbed screens get pretty long, and I see scrollbars that exist on the site today, so kept that functionality as is so users can scroll to see additional detailed result summary content.  The default results summary screen should not need scrollbars though on any viewing width anymore.
- The big piece was getting rid of all the fixed widths, heights, and absolute positioning that don't work well for responsive sites/apps.  The gauges now also resize responsively now based on the viewing width, so that wasn't much of an issue.
- Also centered the NDT test as I think it makes it stand out a little from the page, but if folks like it better left aligned I can change that.

- [ ] One specific note, I'd like to point out as a future task, is we should decide if the front end widget code (HTML & CSS) should live in this repo or in the NDT repo as I see it in both places now.  For example, not sure when we brought in the `ndt-test.css`, but we have that file in this repo and a `style.css` in the [NDT repo](https://github.com/m-lab/ndt/tree/master/HTML5-frontend).  Maintaining this in 2 places probably isn't ideal.  If we decide to maintain that css in the mlab website, then we may want to consider putting it into SASS instead of straight css.  I think this work is already planned, but wanted to point it out for purposes of this PR as well.

Also including some screen caps of the NDT Test at different breakpoints below to easily compare before/after:

resp-ndt

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/183)

### DIFF
--- a/_layouts/ndt-test.html
+++ b/_layouts/ndt-test.html
@@ -69,11 +69,10 @@
       <div id="download" class="test">
         <h2 class="centered">Now testing your <em>download speed</em></h2>
       </div>
-      <canvas id="uploadGauge"></canvas>
-      <div id="rtt">RTT
-        <div id="rttValue"></div>
+      <div class="guage-container">
+          <canvas id="uploadGauge"></canvas>
+          <canvas id="downloadGauge"></canvas>
       </div>
-      <canvas id="downloadGauge"></canvas>
       <div class="locations">
         <div class="local location"></div>
         <div class="remote location">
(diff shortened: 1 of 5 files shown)


critzo:
@shredtechular LGTM. Merging.

PR #196, opened 2016-05-15
Removing suppression of Markdown Linter ruler # 38

shredtechular:
In doing some checking of the Markdown Linter rules that we are suppressing, rule number 38, which is currently being suppressed for spaces inside code span elements is only being flagged on one markdown file which is the `bigquery-schema.md` file.

As you can see in the change below, on line 56 of the `bigquery-schema` file, there was an extra space after `connection_spec.client_ip` and before the closing code span backtick.  I don't see any visual reason for having that space in there, so if we remove the space and re-run the markdown linter, that rule is not flagged anymore and we can remove it.

I put this change on my [fork](http://shredtechular.github.io/m-lab.github.io/data/bq/schema/), so you can compare when that space is removed, there is no visual difference with no space.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/196)

### DIFF
--- a/_linter/markdown-linter-style.rb
+++ b/_linter/markdown-linter-style.rb
@@ -41,8 +41,5 @@ exclude_rule 'MD032'
 ## MD026 - Emphasis used instead of a header
 exclude_rule 'MD036'
 
-## MD038 Spaces inside code span elements
-exclude_rule 'MD038'
-
 ## MD041 First line in file should be a top level header
 exclude_rule 'MD041'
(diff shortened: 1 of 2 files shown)


critzo:
:lgtm:

<details><summary><i>Previously, shredtechular (Andrew Newhouse) wrote…</i></summary>
> #### Removing suppression of Markdown Linter ruler \# 38
> 
> In doing some checking of the Markdown Linter rules that we are suppressing, rule number 38, which is currently being suppressed for spaces inside code span elements is only being flagged on one markdown file which is the `bigquery-schema.md` file.
> 
> 
> 
> As you can see in the change below, on line 56 of the `bigquery-schema` file, there was an extra space after `connection_spec.client_ip` and before the closing code span backtick.  I don't see any visual reason for having that space in there, so if we remove the space and re-run the markdown linter, that rule is not flagged anymore and we can remove it.
> 
> 
> 
> I put this change on my [fork](http://shredtechular.github.io/m-lab.github.io/data/bq/schema/), so you can compare when that space is removed, there is no visual difference with no space.
> 
> 
> 
> 
> 
> </details>


---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/196#-:-KHts8WJn_fwPPZOY3O0:1417088721)*

Issue #266, opened 2017-06-16
update port range in FAQ entry


critzo:
The faq reports that NDT uses ports  49152-65535 but it should be 32768 to 61000. The first range is the IANA recommended ephemeral port range, but many Linux kernels, including ours, apparently use the second range.

Issue #256, opened 2017-05-15
Add FAQ entry "Where are M-Lab Servers hosted? How stable are their connections to the Internet?


critzo:
M-Lab hosts our servers, which we call "pods", in Tier 1 data-centers connected to the internet via 1-10Gbps links. In major metro areas, M-Lab strives to maintain multiple pods, each connected to as many distinct transit providers as are available. 

Once an M-Lab pod is online we make no change in our hosting autonomous system, but we don't control the routes and connectivity of the providers who connect our pods to the internet. That connectivity will change over time as one might expect, as upstream providers peer, upgrade capacity etc. Whenever there is a substantial change to our sites, such as renumbering or relocation, we retire the old site names and metadata, and treat them as brand new sites.

PR #272, opened 2017-07-10
Updated port range per #266 and add FAQ entry per #256

critzo:
This PR adds two updates on the FAQ page. First, it corrects the port ranges used by NDT referenced in #266 and adds an entry to mention where and how we host our servers, addressing #256

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/272)

### DIFF
--- a/_pages/faq.md
+++ b/_pages/faq.md
@@ -20,6 +20,13 @@ The graphic below illustrates the relationship between tests run by consumers an
 
 ![Illustration showing the lifecycle of an M-Lab test]({{ site.baseurl }}/images/about/lifecycle-of-mlab-test.png)
 
+## Where are M-Lab Servers hosted?
+{:.no_toc}
+
+M-Lab hosts our servers, which we call "pods", in Tier 1 data-centers, connected to the internet with links between 1-10 Gbps. In major metro areas, M-Lab strives to maintain multiple pods, each connected to as many distinct transit providers as possible.
+
+Once an M-Lab pod is online we make no changes in our systems for the life of each pod, but we don't control the routes and connectivity of the providers who connect our pods to the internet. That connectivity will change over time as one might expect, as upstream providers peer, upgrade capacity etc. Whenever there is a substantial change to our sites, such as renumbering of IP addresses, we retire the pod, retain its naming and metadata in our systems, and treat the pod as brand new.
+
 # M-Lab's Tests and Data
 
 ## What measurement tests can I run?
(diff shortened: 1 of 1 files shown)


_pages/faq.md
```diff
@@ -20,6 +20,13 @@ The graphic below illustrates the relationship between tests run by consumers an
 
 ![Illustration showing the lifecycle of an M-Lab test]({{ site.baseurl }}/images/about/lifecycle-of-mlab-test.png)
 
+## Where are M-Lab Servers hosted?
+{:.no_toc}
+
+M-Lab hosts our servers, which we call "pods", in Tier 1 data-centers, connected to the internet with links between 1-10 Gbps. In major metro areas, M-Lab strives to maintain multiple pods, each connected to as many distinct transit providers as possible.
```
georgiamoon: Should we add a link here to the infrastructure page?

_pages/faq.md
```diff
@@ -20,6 +20,13 @@ The graphic below illustrates the relationship between tests run by consumers an
...
+
+Once an M-Lab pod is online we make no changes in our systems for the life of each pod, but we don't control the routes and connectivity of the providers who connect our pods to the internet. That connectivity will change over time as one might expect, as upstream providers peer, upgrade capacity etc. Whenever there is a substantial change to our sites, such as renumbering of IP addresses, we retire the pod, retain its naming and metadata in our systems, and treat the pod as brand new.
```
georgiamoon: This sentence `Once an M-Lab pod is online we make no changes in our systems for the life of each pod, but we don't control the routes and connectivity of the providers who connect our pods to the internet.` is a little confusing. Maybe rephrase?

`as one might expect` -- remove

critzo: A rephrase:
`Once an M-Lab pod is online we make no changes in our systems for the life of the pod.`  --just dropping the part about routes to the pod...

_pages/faq.md
critzo: I was thinking that, but then didn't. Do you think we should add a menu link to it as well? Maybe under 'About'?

critzo:
@georgiamoon I made an update based on your feedback.

_pages/faq.md
georgiamoon: I think just a link here is fine. I don't think under About makes sense, but I'd want to revisit the site architecture notes that we have, where we had a plan for where it goes. Let's keep that as out of scope of this PR

georgiamoon:
@critzo should be good with a link to infrastructure in that first paragraph. LGTM.

_pages/faq.md
critzo: sgtm

PR #302, opened 2017-08-28
Update privacy.md

pboothe:
We said "public facing IP address of the client" one place, and "user's IP address" another. The first is more correct, so I have left it in and removed the "user's" modifier on the later use of "IP address".

I have no idea why that last line is shown as modified. I suspect it lacks a trailing newline, and the webform has helpfully added one.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/302)


critzo:
LGTM @pboothe

PR #329, opened 2017-12-01
update site page redirects for previously moved pages

critzo:
This PR adds page redirects using a jekyll redirect template, updates the readme, and removes old AWS-based redirects.
- redirects were being accomplished using a root level folder containing a zero byte index file, which then had 301 redirect metadata applied to it in AWS. this stopped working for some unknown reason
- new implementation uses a jekyll template
- each redirected page begins with `301-` and contains yaml variables to both continue providing the original link and to redirect the user to the the URL. 

Previewable here: https://critzo.github.io/m-lab.github.io/

@georgiamoon  this PR should be easier to parse-- one commit, although a bunch of files added to replace the ones that are being removed.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/329)

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,17 +44,16 @@ This site enables HTML Compression for optimizing performance.  If it is desired
 | images | Contains all the image files for the site. |
 | publications | Contains all the pdfs and docs that the site links to. |
 
-## 301 Redirects
+## Page Redirects
 
 For page redirects, M-Lab contributors should:
 
-* create the new page in it's new location, using the _permalink_ part of the page's YAML front matter.
+* Create the new page in it's new location, using the _permalink_ part of the page's YAML front matter.
   * For example the page `_pages/gsc.md` used to be served at the URL: `https://www.measurementlab.net/data/gcs/`
   * It was "moved" to a new URL by changing the _permalink_ front matter from `permalink: /data/gcs/` to `permalink: /data/docs/gcs/`.
-* prepare empty "stub" folder(s) and file(s) to use for the redirect.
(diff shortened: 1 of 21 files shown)


georgiamoon:
Question: this doesn't seem to interact well with the `{{ site url }}` in that the redirects on the critzo.github.io version go to measurementlab.net. Is there a way to have it work with the `{{ site url }}` preview mode or not really?

critzo:
There is a way to test the site when you run it on your [local machine](https://github.com/m-lab/m-lab.github.io/blob/master/README.md#local-development) by using the `--baseurl` flag when you ask jekyll to serve it. To do that on the github build on a fork would require changing the URL in the site's main [yaml config file](https://github.com/m-lab/m-lab.github.io/blob/master/_config.yml#L10) and then undoing that before merging. For this reason I usually don't do that. Let me know if you want to do this to confirm the changes.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/m-lab/m-lab.github.io/329#-:-L-mA-0stiFbC8EudgmB:b3h1eig)*

georgiamoon:
Chatting with @critzo now, we're going to push this change out to address the current redirects issue, and look into the local testing issues next. The concern is that it won't be possible to test a redirect to a new page that doesn't exist yet, e.g. when we're creating new redirects in the future, because the page will not exist yet on measurementlab.net, so ideally we want a way to test locally, or we'll need a process where we publish the new page, and then create a the redirect in a second pull request.

PR #339, opened 2018-01-18
Add Georgia Bullen to steering committee members

critzo:
@georgiamoon If you have an alternate photo you'd like me to use, please let me know.

previewable here: https://critzo.github.io/m-lab.github.io/who/

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/339)


georgiamoon:
LGTM. 

Note for later -- when previewing the site, the links to the external sites are pre-pended with the site URL, but that doesn't happen on the live site for some reason. (or we should confirm that after pushing this change).

PR #362, opened 2018-04-09
Add blog post about murakami

rschulman:
This PR adds a blog post (which has already been reviewed) about the new `murakami` test platform. It also adds a month entry for `2018-04`.


critzo:
LGTM @rschulman

PR #387, opened 2018-07-30
clarify current locations of PT and SS tables

critzo:
This is an update to the last PR updating the schema page, but adds reference to the current publication location for Paris Traceroute and Sidestream tables. @georgiamoon could you please review?

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/387)


georgiamoon:
Any reason to not just directly link these to the right views?

PR #437, opened 2018-12-20
Remove duplicated passage in Definitions:RTT

apjanke:
Looks like a copy-paste error resulted in the text for "Download Throughput" being repeated under the "Round Trip Time" section in Definitions.

This PR removes the duplicated text.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/437)


critzo:
Thanks for finding this @apjanke. I don't think we have duplication here, as much as we have some poorly worded text that needs some revision and formatting. Logged the issue in #452.

PR #441, opened 2018-12-20
Forbid non-HTTPS www.measurementlab.net URLs

### LINKED ISSUE #166: Build check: Forbid "http://measurementlab.net"
We should add a build check that checks Jekyll files for the strings `http://measurementlab.net` or `http://www.measurementlab.net` and flags the user to convert these to `{{ site.baseurl }}`

apjanke:
Here's a custom markdownlint linter to forbid non-HTTPS links to `www.measurementlab.net`, as requested by #166.

Includes a fix for a bad link it caught.

Closes #166.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/441)

### DIFF
--- a/_linter/markdown-linter-rules.rb
+++ b/_linter/markdown-linter-rules.rb
@@ -1,0 +1,22 @@
+# Custom linter rules for M-Lab
+#
+# These rule names use the prefix "ML" for "Measurement Lab"
+#  ----------------------------------------------------------------------------
+
+
+rule "ML001", "Forbid non-HTTPS www.measurementlab.net links" do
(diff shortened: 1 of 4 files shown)


critzo:
Thanks for this contribution @apjanke , but for various reasons we don't really need this custom linter rule anymore. Links to internal site pages should not use the full URL, but instead use the Liquid variable `{{ site.baseurl }}`. For example `[link text]({{ site.baseurl }}/path/file.pdf)`. The site hosting now has redirects so all pages redirect to `https://www.measurementlab.net/<pagerequest>` so this should no longer be needed.

PR #442, opened 2018-12-20
Fix markdownlint MD036 errors

apjanke:
This PR fixes the MD036 markdownlint errors by replacing emphasis with lower-level headers, and re-enables the rule in the style config.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/442)


apjanke:
I can't tell why the Travis CI check is failing here. It looks almost like it only pulled the second commit in this PR?

apjanke:
Ah, I see: it looks like it's failing because the Travis CI checks are using an older version of markdownlint, and I'm using markdownlint 0.5.0. The older versions are getting more positives for "emphasis used instead of header".

critzo:
@apjanke This PR changes a bunch of text from emphasis to headings, but we intended on these elements to not be headings. We'll likely do a content/theming review and we could return to this later.

PR #440, opened 2018-12-20
CONTRIBUTING: Add additional dependencies to setup instructions

apjanke:
Fixes #430.

Running the website locally requires Bundler, Jupyter, and doing the pre-commit checks requires markdownlint and Node.js, but they're not included in the local dev environment setup instructions. This PR adds them.

An alternative to listing markdownlint might be to replace the plain `mdl` command in _tests/travis-checks with `bundle exec mdl`. (https://github.com/m-lab/m-lab.github.io/pull/443 does this.)

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/440)

PR #433, opened 2018-12-19
Local dev env: add explicit empty string arg to --baseurl

apjanke:
In the "local dev environment" part of CONTRIBUTING:

I think you need to pass an empty string as the argument to `jekyll serve --baseurl`, not just omit the argument.

Using the current instructions:

```
$ bundle exec jekyll serve --baseurl
bundler: failed to load command: jekyll (/usr/local/lib/ruby/gems/2.5.0/bin/jekyll)
OptionParser::MissingArgument: missing argument: --baseurl
  /usr/local/lib/ruby/gems/2.5.0/gems/mercenary-0.3.6/lib/mercenary/program.rb:31:in `go'
  /usr/local/lib/ruby/gems/2.5.0/gems/mercenary-0.3.6/lib/mercenary.rb:19:in `program'
  /usr/local/lib/ruby/gems/2.5.0/gems/jekyll-3.6.2/exe/jekyll:15:in `<top (required)>'
  /usr/local/lib/ruby/gems/2.5.0/bin/jekyll:23:in `load'
  /usr/local/lib/ruby/gems/2.5.0/bin/jekyll:23:in `<top (required)>'
```

Using an explicit `""` argument:

```
$ bundle exec jekyll serve --baseurl ""
Configuration file: /Users/janke/local/repos/m-lab.github.io/_config.yml
       Deprecation: The 'gems' configuration option has been renamed to 'plugins'. Please update your config file accordingly.
            Source: /Users/janke/local/repos/m-lab.github.io
       Destination: /Users/janke/local/repos/m-lab.github.io/_site
 Incremental build: disabled. Enable with --incremental
      Generating...
```

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/m-lab.github.io/433)

PR #453, opened 2019-02-11
update dependencies in contributing.md; move Neubot to retired status

critzo:
This PR adds some content suggested in #440 and #433, updates the Neubot test status and removes links to resources that NEXA no longer hosts, since Neubot is now decommissioned. @georgiamoon  PTAL?

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/website/453)


georgiamoon:
So the tricky bit here is that the Neubot client is deprecated, but the server isn't. So I'm not sure we actually want to represent Neubot as retired, since we aren't removing the slice right?

cc @bassosimone

critzo:
> So the tricky bit here is that the Neubot client is deprecated, but the server isn't. So I'm not sure we actually want to represent Neubot as retired, since we aren't removing the slice right?
> 
> cc @bassosimone

I believe the plan is that @bassosimone will apply to host a new test, DASH, on the new Docker-based platform, and that when the platform is migrated, the Neubot server will be retired. Perhaps this PR was premature. I found that the client download links and other URLs from Nexa's site were not resolving, so I took the extra step of anticipating the retirement. @georgiamoon would removing dead links and "retired" language/placement be ok to move this PR forward?

critzo:
@georgiamoon I've changed the Neubot references to indicate only the client has been retired, and that the server will continue to operate on M-Lab until the platform transition. PTAL?

critzo:
Thanks @bassosimone !

PR #687, opened 2022-03-16
Update dependencies and add Gemfile.lock

robertodauria:
Some of the Ruby dependencies stopped working correctly with the version of Ruby included in Ubuntu Focal (20.04) This change has let them update to the latest version. 

Also, the current best practices for Ruby projects seem to be to always commit Gemfile.lock, so I've removed it from .gitignore and committed it.

---
This change is [Reviewable](https://reviewable.io/reviews/m-lab/website/687)


laiyi-ohlsen (approved):
LGTM, thank you!

PR #756, opened 2023-07-10
publish blog post about july 2023 office hours

laiyi-ohlsen:


KatataK:
LGTM. Haven't used blog posts as a format for regular meetings. Do you think we should do the same for community calls? Or is the idea to have a few blog posts at the outset introducing office hours and once well established move it to a different location on the website(support? contact? get involved? Ask Questions?)

KatataK:
:lgtm:

laiyi-ohlsen:
The thought here was to have the Learn page be the long-term resource for the office hours and just do a blog post when one is upcoming as long as it makes sense to. And totally, this is what I was starting to talk through at standup a bit, I think we can think about how to organize our events & calls cohesively. 

Thanks for the lgtm, looks like it still needs the formal approval.

KatataK (approved):

PR #759, opened 2023-08-14
Add text labels to metros with a single site

stephen-soltesz:
The default mode for mapbox status map is to leave unclustered metros without a label. Without a label, these locations are ambiguous -- are they M-lab sites or city locations on the underlying geography? As we migrate away from mlab-ns as the canonical map of server locations, it is helpful to improve the clarity of this status map.

This change makes several small configuration updates to:
* add a label of "1" on unclustered metros (where there is a single site)
* recenter the map so we can zoom in slight so more metros show in the default view
* increases the default circle size slightly: radius 7 (for metros with 1-2 sites), 14 (for metros 3-5 sites), 21 greater or equal to 6.

An example of this change is below and on the sandbox website.

Original view:
Screen Shot 2023-08-14 at 4 53 30 PM

Updated view:
Screen Shot 2023-08-14 at 4 47 00 PM

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/website/759)


laiyi-ohlsen (approved):
Nice, good catch. LGTM>

PR #805, opened 2024-07-11
Fixe Analytics GA4 tag inclusion for measurementlab.net

nkinkade:
The way I had done it previously was resulting in nested <script> tags, which was breakting things. This commit splits out the loading of gtag/js into its own script tag, and reduces ga.js to only javascript.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/website/805)


laiyi-ohlsen (approved):
LGTM

PR #850, opened 2025-12-03
Update NDT data documentation

nkinkade:
This PR attempts to update the NDT data documentation:

* Updates currently recommended view from "unified_*" to "ndt7_union"
* Adds a new section describing the difference between "M-Lab-managed" data and "Host-Managed" data
* Removes the ugly Markdown table with description of views in favor of a cleaner (in my opinion) header-based approach.
* Modifies various view descriptions, possibly adding supporting links.

http://website.mlab-sandbox.measurementlab.net/tests/ndt/index.html#m-lab-managed-vs-host-managed-ndt-data

@mattmathis: What do you think?

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/website/850)


nkinkade:
I bolded both of your suggested phrases. Once the build completes I'll go ahead and merge this so that we can at least get a slightly better set of documentation publishes. From there, we can iterate as things changes. Thanks.

PR #877, opened 2026-06-16
static data portal for website, starting with IQB data

jduckles:
Porting @bassosimone's IQB data portal to the main M-Lab website as-is for now. 

This should work for now, but when we get the new website going I'll try to develop an approach that creates a landing page at /data, but for now this should make iqb data available at https://measurementlab.net/data/iqb

Note: I suggest maybe the `manifest.json` be written to the bucket and the bucket have CORS 'Access-Control-Allow-Origin: *' set so that people can build apps off the data directly and we avoid having to write a new manifest to GitHub every time the data are updated, but this will get it working for now. 

Note2: Additionally, I wonder why we're not using just JSON so that JS apps can be build directly off this data. They seem unusually small to be using parquet files for them. Gzipped JSON and Parquet are within a few kb of each other in size. 

Note 3: Trying to use this data with each file named `data.parquet` is a bit frustrating as well, so we may want to think about how researchers / analysts might benefit from a bit of metadata in the filename not just in the path going forward. 

Also @bassosimone where are the scripts running that dump these data?

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/website/877)


sermpezis:
On note 2: 
Some data files are large, e.g., for `2025-12 – 2026-01 | uploads | country + city + asn`, data in .parquet format is ~18MB.

data/iqb/index.html
```diff
@@ -0,0 +1,282 @@
...
+    // Wire up filter change events
+    var selects = document.querySelectorAll("#filters select");
+    for (var k = 0; k < selects.length; k++) {
+      selects[k].addEventListener("change", applyFilters);
+    }
+
+    applyFilters();
+  }
+
+  fetch("manifest.json")
```
bassosimone, line 271:
```suggestion
  // TODO(bassosimone): we are currently embedding a manifest. Pros: we cannot break
  // this website by changing IQB in incompatible ways. Cons: the content may become
  // stale. For now this is fine, but we should see to improve upon this in the future.
  fetch("manifest.json")
```

bassosimone (approved):
Please, apply my suggestion. (I typed the code from the web, so worth double checking my suggestion does not break anything.) After this, it's fine to merge. Thanks!

PR #881, opened 2026-06-23
Change sort order and updating directory and language iqb->stats 

jduckles:
This PR changes the name of the dataset here from `iqb` to `stats` and changes the sort order so the most recent data is first on the page.

- - -
This change is [Reviewable](https://reviewable.io/reviews/m-lab/website/881)


jduckles:
Hey @sermpezis please review and merge before Thursday CoP call so we can use /data/stats not ~~/data/iqb~~ when sharing the monthly stats data.

sermpezis (approved):

